PR #3, opened 2024-08-16
Add Default Install For RHDH Operator

michael-valdron:
This PR adds the portion of the Helm Chart that facilitates the deployment of the RHDH Operator. This also creates an instance of RHDH in the target namespace.

Portions of this chart were referenced from RHTAP: https://github.com/redhat-appstudio/rhtap-installer/tree/main/chart

## Additional Changes

- https://github.com/redhat-ai-dev/ai-rhdh-installer/commit/83c9bbda6d29835c628f647a8d26f3e621e80263 - fixes argocd instance deployment to use target namespace instead of default namespace
- https://github.com/redhat-ai-dev/ai-rhdh-installer/commit/3f168b4b952ef4402c4c3bfaf1be5ae9910d621d - adds unconfigure batch job to delete resources on `helm uninstall`
- https://github.com/redhat-ai-dev/ai-rhdh-installer/commit/bdd04c76e830658f9bba4e45dd4cdc91c2b7ddb5 - amend instructions to not use default namespace

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,9 @@ This helm chart installs and configures the following operators:
 
 |       Product       |      Installation       |                                                                                                      Configuration                                                                                                       |
 | :-----------------: | :---------------------: | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------: |
-|  OpenShift GitOps   | Operator `Subscription` |                                    Controlled by the values YAML file. If a subscription already exists, the installation will not modify it. A new instance ArgoCD will be created.                                     |
+|  OpenShift GitOps   | Operator `Subscription` |                                    Controlled by the values YAML file. If a subscription already exists, the installation will not modify it. A new instance of ArgoCD will be created.                                     |
 | OpenShift Pipelines | Operator `Subscription` | Controlled by the values YAML file. If a subscription already exists, the installation will not modify it. In all cases, the TektonConfig will be modified to enable Tekton Chains and the signing secret will be setup. |
+| Red Hat Developer Hub | Operator `Subscription` | Controlled by the values YAML file. If a subscription already exists, the installation will not modify it. A new instance of RHDH will be created. |
 
 **Note**: If a subscription for an operator already exists, the installation will not tamper with it.
 
--- a/README.md
+++ b/README.md
@@ -26,15 +26,19 @@ This helm chart installs and configures the following operators:
 
 ### Install
 
-Run `helm upgrade --install <release-name> <path-to-chart>` to deploy default installations of the necessary operators.
+Run `helm upgrade --install <release-name> <path-to-chart> --namespace <namespace> --create-namespace` to deploy default installations of the necessary operators.
 
 #### Example
 
-`helm upgrade --install setup-default-operators ./chart`
+`helm upgrade --install ai-rhdh ./chart --namespace rhdh --create-namespace`
 
 ### Uninstall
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -26,15 +27,19 @@ This helm chart installs and configures the following operators:
...
+`helm upgrade --install ai-rhdh ./chart --namespace rhdh --create-namespace`
 
 ### Uninstall
 
-Since for the default installation the `ServiceAccount` is being deployed to the `default` namespace with admin permissions you are unable to remove it during a `helm uninstall`. You first need to remove the ServiceAccount manually before running `helm uninstall`
+`helm uninstall <release-name> --namespace <namespace>`
+
+#### Default Namespace
+
+Since for the default installation the ServiceAccount is being deployed to the `default` namespace with admin permissions you are unable to remove it during a `helm uninstall`. You first need to remove the ServiceAccount manually before running `helm uninstall`
```
maysunfaisal: Just curious, why cant we have default installation be something other than the `default` namespace?

michael-valdron: > Just curious, why cant we have default installation be something other than the `default` namespace?

@maysunfaisal Yeah should indicate incompatibility instead, just updated this section to indicate incompatibility with the default namespace.

maysunfaisal (approved):

chart/templates/developer-hub/includes/_backstage.tpl
```diff
@@ -0,0 +1,15 @@
+{{ define "rhdh.include.backstage" }}
+---
+apiVersion: rhdh.redhat.com/v1alpha1
+kind: Backstage
+metadata:
+  name: ai-rhdh-developer-hub
```
thepetk: Might be better if the name here is `ai-rh-developer-hub`?

README.md
```diff
@@ -26,18 +27,16 @@ This helm chart installs and configures the following operators:
 
 ### Install
 
-Run `helm upgrade --install <release-name> <path-to-chart>` to deploy default installations of the necessary operators.
+Run `helm upgrade --install <release-name> <path-to-chart> --namespace <namespace> --create-namespace` to deploy default installations of the necessary operators.
 
 #### Example
 
-`helm upgrade --install setup-default-operators ./chart`
+`helm upgrade --install ai-rhdh ./chart --namespace rhdh --create-namespace`
```
thepetk: sticking to the name `ai-rhdh` shouldn't we name the namespace something related with `ai-rhdh`?

thepetk (commented):
I've tested on a local cluster the updates:

* I was able to install RHDH and the other listed operators.
* The only issue I had was to login through github, but I think this is out of the scope of the issue.

I left some minor comments for the names used. Might be good to follow the ai-rhdh or ai-rh-developer-hub terms everywhere. If not possible, feel free to ignore :)

README.md
Jdubrick: I think `rhdh` is okay for the namespace since this will just be a normal install of RHDH that is compatible with AI software templates even though it isn't required you use them.

Jdubrick (approved):
lgtm after the comment from Fanis is addressed

README.md
michael-valdron: I agree with @Jdubrick, I don't think its too important what a user uses here but I will change it still to use consistent naming under the example.

thepetk: Nice, yeah if we're going to stick to the normal installation of RHDH is fine to keep the `rhdh` name

thepetk (approved):
/lgtm

PR #6, opened 2024-09-09
Add GitOps/ArgoCD Configuration Script

Jdubrick:
### What does this PR do?:

This PR allows a user to configure RHDH with ArgoCD/GitOps. Since the scope of this PR and linked issue is only for GitOps/ArgoCD it is assumed that a RHDH instance has already been configured to the point of being able to login (i.e. setup the github app and secrets necessary for login).

Once this PR is completed and through review I will record some demos outlining the different methods of configuration.

### Which issue(s) this PR fixes:

https://issues.redhat.com/browse/DEVAI-69

### PR acceptance criteria:
Testing and documentation do not need to be complete in order for this PR to be approved. We just need to ensure tracking issues are opened and linked to this PR, if they are not in the PR scope due to various constraints.

- [x] Tested and Verified

  

- [x] Documentation (READMEs, Product Docs, Blogs, Education Modules, etc.)

   

### How to test changes / Special notes to the reviewer:
In order to test these changes you first need a RHDH instance setup and configured. You can easily setup a RHDH instance by using the `Helm` install that is apart of this repo since the config scripts are separate. Once a RHDH instance has been created you can reference [this doc](https://docs.google.com/document/d/1z9yXX2KS-eUJVLIUFiXtuzpKmKzw5aUYWTcXc44ZwF0/edit?usp=sharing) to quickly set it up manually.

Once RHDH is ready you can follow the instructions in the changes I made to the README to get everything up and running.

PR #7, opened 2024-09-13
Add Pipelines/Tekton Configuration Script

michael-valdron:
### What does this PR do?:

This PR allows a user to configure RHDH with Pipelines/Tekton.

**Additional changes**:
- Plugins are no longer separated into different ConfigMaps and are instead merged into a target dynamic plugins ConfigMap tied to the RHDH deployment, [GitOps config doc](https://github.com/michael-valdron/ai-rhdh-installer/blob/feat/pipelines-config/docs/GITOPS-CONFIG.md) and [Pipelines config doc](https://github.com/michael-valdron/ai-rhdh-installer/blob/feat/pipelines-config/docs/PIPELINES-CONFIG.md) are updated to reflect this change to both scripts
- [Private env file template](https://github.com/michael-valdron/ai-rhdh-installer/blob/feat/pipelines-config/default-private.env) provided as a starting point to set private environment variables to export for the scripts, `private*` files are now ignored
- `.<component>.enabled` now toggles whether the component is enabled

resolves https://issues.redhat.com/browse/DEVAI-31

### PR acceptance criteria:

- [x] Tested and Verified

  

- [X] Documentation (READMEs, Product Docs, Blogs, Education Modules, etc.)

   

You'll need an RHDH instance has already been configured to the point of being able to login (i.e. setup the github app and secrets necessary for login). This is to give you proper permissions under RHDH to verify the plugins are installed, guest mode does not give enough permissions to allow for this. Unlike the GitOps/ArgoCD Configuration Script (#6), the Pipelines/Tekton Configuration Script can work with RHDH default setup.

Follow the steps to setup developer hub with GitHub integration and login support:
1. Create the following ConfigMap to use for the extra app config with a Secret that has all the referenced env vars in it:
    - **Note**: `catalog.locations[0].target` is set to this [link](https://github.com/Jdubrick/ai-lab-template/blob/e0ffa0206b3f666b65bb68f75b00dbe9a4a4c2e6/all.yaml) in order to test with the proper new pipelines definition referenced, changes not in place yet in the template repository.
```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: app-config-rhdh-extra
  namespace: ai-rhdh
data:
  app-config.extra.yaml: |
    app:
      title: Red Hat Developer Hub
      baseUrl: <developer_hub_route_url>
    integrations:
      github:
        - host: github.com
          apps:
            - appId: ${GITHUB_APP_APP_ID}
              clientId: ${GITHUB_APP_CLIENT_ID}
              clientSecret: ${GITHUB_APP_CLIENT_SECRET}
              webhookUrl: ${GITHUB_APP_WEBHOOK_URL}
              webhookSecret: ${GITHUB_APP_WEBHOOK_SECRET}
              privateKey: |
                ${GITHUB_APP_PRIVATE_KEY}
    auth:
      # see https://backstage.io/docs/auth/ to learn about auth providers
      environment: development
      providers:
        github:
          development:
            clientId: ${GITHUB_APP_CLIENT_ID}
            clientSecret: ${GITHUB_APP_CLIENT_SECRET}
    catalog:
      providers:
        github:
          myorg:
            organization: '${GITHUB_ORG}'
          schedule:
            # supports cron, ISO duration, "human duration" (used below)
            frequency: { minutes: 30}
            # supports ISO duration, "human duration (used below)
            timeout: { minutes: 3}
            initialDelay: { seconds: 15}
      import:
        entityFilename: catalog-info.yaml
        pullRequestBranchName: rhdh-integration
      rules:
        - allow: [Component, System, Group, Resource, Location, Template, API]
      locations:
        - type: url
          target: https://github.com/Jdubrick/ai-lab-template/blob/e0ffa0206b3f666b65bb68f75b00dbe9a4a4c2e6/all.yaml

    proxy:
      endpoints:
        '/quay/api':
          target: 'https://quay.io/'
          headers:
            X-Requested-With: 'XMLHttpRequest'
          
            Authorization: 'Bearer ${QUAY_API_TOKEN}'
          
          changeOrigin: true
          # Change to "false" in case of using self hosted quay instance with a self-signed certificate
          secure: true
    quay:
      # The UI url for Quay, used to generate the link to Quay
      uiUrl: 'https://quay.io/'
    techdocs:
      builder: 'local'
      generator:
        runIn: 'local'
      publisher:
        type: 'local'
```
2. Patch the default app config that's create with the RHDH instance to include the following to fix the `redirect-uri` error, `backend` parent should already exist:
```yaml
backend:
  ...
  baseUrl: <developer_hub_route_url>
  cors:
    origin: <developer_hub_route_url>
```
3. Update the rhdh container deployment spec in the developer view of the OpenShift Console to reference the Secret in the env vars and mount the ConfigMap file as a volume
    1. Under `.spec.template.spec.volumes`, add the following:
    ```yaml
    - name: app-config-rhdh-extra
      configMap:
        name: app-config-rhdh-extra
        defaultMode: 420
        optional: false
    ```
    2. Under `.spec.template.spec.containers`, find the container entry named `backstage-backend`, should be the only one, and perform the following changes under this entry
        1. Under `fromEnv` add the following entry:
        ```yaml
        - secretRef:
            name: <name-of-the-secret-mentioned-earlier>
        ```
        2. Under `volumeMounts` add the following entry:
        ```yaml
        - name: app-config-rhdh-extra
          readOnly: true
          mountPath: /opt/app-root/src/app-config.extra.yaml
          subPath: app-config.extra.yaml
        ```
        3. Under `args` add the following two entries:
        ```yaml
        - '--config'
        - /opt/app-root/src/app-config.extra.yaml
        ```
4. Save the edited deployment spec, after the pod restarts, GitHub login should work


michael-valdron:
/do-not-merge work-in-progress

Marking as ready for review now as all content related to the pipelines additions are ready for review, still blocking from merging as I'm just waiting on setups to provide updated screenshots for the GitOps doc plugins revision.

Should be done shortly.

FYI @Jdubrick @maysunfaisal

michael-valdron:
> /do-not-merge work-in-progress
> 
> Marking as ready for review now as all content related to the pipelines additions are ready for review, still blocking from merging as I'm just waiting on setups to provide updated screenshots for the GitOps doc plugins revision.
> 
> Should be done shortly.
> 
> FYI @Jdubrick @maysunfaisal

Done.

docs/GITOPS-CONFIG.md
```diff
@@ -20,31 +20,43 @@ If you installed using the [`ai-rhdh-installer`](../README.md#install) all that
 ### AI-RHDH-Installer: Manual Configuration
 
 #### Step 1: ConfigMaps and Secrets
-In your chosen namespace you should apply the following ConfigMaps:
-1.  [`argocd-config.yaml` ](../resources/argocd-config.yaml)
-2.  [`argocd-plugins.yaml`](../resources/argocd-plugins.yaml)
+In your chosen namespace you should apply the `argocd-config` ConfigMap under [`argocd-config.yaml` ](../resources/argocd-config.yaml).
 
 As part of the `ai-rhdh-installer` a secret was created in your desired namespace with the name `rhdh-argocd-secret`, keep note of this Secret as well as the two ConfigMaps applied above.
```
thepetk, line 26:
```suggestion
As part of the `ai-rhdh-installer` a secret was created in your desired namespace with the name `rhdh-argocd-secret`, keep note of this Secret as well as the ConfigMap applied above.
```

thepetk: Here I think it gets confusing. Installing with `ai-rhdh-installer` is mentioned as pre-requisite for the script configuration. Reading this sentence makes me think it should be for both right?

docs/GITOPS-CONFIG.md
```diff
@@ -20,31 +20,43 @@ If you installed using the [`ai-rhdh-installer`](../README.md#install) all that
...
-Edit the associated `yaml` file to include `argocd-plugins.yaml` under the `includes` section:
+Edit the associated `yaml` file to include the contents of the [ArgoCD plugins list](../dynamic-plugins/argocd-plugins.yaml) under the `plugins` section:
 
 ![Dynamic Plugins Example 2](../assets/dynamic-plugins-example-2.png)
 
-#### Step 3: Updating RHDH Deployment
-Now that all of the required ConfigMaps and Secrets are apart of the namespace, as well as adding the plugins we must edit the RHDH Deployment to properly reference these items.
-
-Since we are referencing `argocd-plugins.yaml` in our dynamic-plugins ConfigMap we first must add `argocd-plugins.yaml` as a `volumeMount` to the `initContainers` field in the RHDH Deployment under `.spec.template.spec.initContainers.volumeMounts`
+Alternatively, we can use this series of commands to perform the same task with `kubectl` and `yq` using the [`argocd-plugins.yaml`](../dynamic-plugins/argocd-plugins.yaml):
```
thepetk, line 41: Might be more clear here if we have 2 sub-sections:

* Edit the associated `yaml` file to include..
* OR Alternatively, we can use this series...

docs/GITOPS-CONFIG.md
```diff
@@ -20,31 +20,43 @@ If you installed using the [`ai-rhdh-installer`](../README.md#install) all that
...
+
+1. Fetch the dynamic plugins ConfigMap and save the `dynamic-plugins.yaml` content within to a temp file
+    ```sh
+    kubectl get configmap backstage-dynamic-plugins-ai-rh-developer-hub -n ai-rhdh -o yaml | yq '.data["dynamic-plugins.yaml"]' > temp-dynamic-plugins.yaml
```
thepetk: The namespace might defer from `ai-rhdh` right? So it might be better here to include something like `<your-namespace>`?

docs/GITOPS-CONFIG.md
```diff
@@ -20,31 +20,43 @@ If you installed using the [`ai-rhdh-installer`](../README.md#install) all that
...
+    ```sh
+    yq -i ".plugins += $(yq '.plugins' ./dynamic-plugins/argocd-plugins.yaml -M -o json) | .plugins |= unique_by(.package)" temp-dynamic-plugins.yaml
+    ```
+3. Patch the dynamic plugins ConfigMap with the updated content in the temp file
+    ```sh
+    kubectl patch configmap backstage-dynamic-plugins-ai-rh-developer-hub -n ai-rhdh \
+    --type='merge' \
+    -p="{\"data\":{\"dynamic-plugins.yaml\":\"$(echo "$(cat temp-dynamic-plugins.yaml)" | sed 's/"/\\"/g' | sed 's/$/\\n/g' | tr -d '\n')\"}}"
+    ```
+4. Dynamic plugins should be updated with the [ArgoCD plugins list](../dynamic-plugins/argocd-plugins.yaml) with a pod update triggered and you may remove the temp file at this point
```
thepetk:
```suggestion
4. Dynamic plugins should be updated to include the [ArgoCD plugins list](../dynamic-plugins/argocd-plugins.yaml) with a pod update triggered and you may remove the temp file at this point
```

thepetk: Just to avoid having "with" two times.

docs/GITOPS-CONFIG.md
```diff
@@ -20,31 +20,43 @@ If you installed using the [`ai-rhdh-installer`](../README.md#install) all that
...
 
-![ArgoCD Plugins Addition](../assets/argocd-plugins-example.png)
+#### Step 3: Updating RHDH Deployment
+Now that all of the required ConfigMaps and Secrets are apart of the namespace, we must edit the RHDH Deployment to properly reference these items.
```
thepetk: Might be my english here, just not quite sure if "are apart of" == "are part of" 
```suggestion
Now that all of the required ConfigMaps and Secrets are part of the namespace, we must edit the RHDH Deployment to properly reference these items.
```

docs/GITOPS-CONFIG.md
```diff
@@ -111,9 +121,9 @@ You will also need to create a Secret in your chosen namespace with the followin
 4. `ARGOCD_API_TOKEN`
 
 #### Step 3: Updating Plugins
-You will follow the same steps as [step 2 for the ai-rhdh-installer](#ai-rhdh-installer-script-configuration)
+You will follow the same steps as [step 2 for the ai-rhdh-installer](#step-2-updating-plugins).
```
thepetk:
```suggestion
Follow the same steps as [step 2 for the ai-rhdh-installer](#step-2-updating-plugins).
```

docs/GITOPS-CONFIG.md
```diff
@@ -111,9 +121,9 @@ You will also need to create a Secret in your chosen namespace with the followin
...
 
 #### Step 4: Updating RHDH Deployment
-Once you have applied the ConfigMaps and Secrets to your cluster and the necessary namespace you can now follow the same steps in [step 3 for the ai-rhdh-installer](#ai-rhdh-installer-manual-configuration). Every step will be identical if your RHDH instance was created using the `Red Hat Developer Hub Operator`, however, if you installed RHDH using the `Helm Chart` you may find that `.spec.template.spec.containers.envFrom` does not exist in the Deployment yaml. If this is the case you can simply add that field and the necessary secret information.
+Once you have applied the ConfigMaps and Secrets to your cluster and the necessary namespace you can now follow the same steps in [step 3 for the ai-rhdh-installer](#step-3-updating-rhdh-deployment). Every step will be identical if your RHDH instance was created using the `Red Hat Developer Hub Operator`, however, if you installed RHDH using the `Helm Chart` you may find that `.spec.template.spec.containers.envFrom` does not exist in the Deployment yaml. If this is the case you can simply add that field and the necessary secret information.
```
thepetk:
```suggestion
Once you have applied the ConfigMaps and Secrets to your cluster and the chosen namespace you can now follow the same steps in [step 3 for the ai-rhdh-installer](#step-3-updating-rhdh-deployment). Every step will be identical if your RHDH instance was created using the `Red Hat Developer Hub Operator`, however, if you installed RHDH using the `Helm Chart` you may find that `.spec.template.spec.containers.envFrom` does not exist in the Deployment yaml. If this is the case you can simply add that field and the necessary secret information.
```

thepetk: Might be better also here to be more detailed:

* `If this is the case you can simply add that field and the necessary secret information` 

I think maybe add this info as troubleshooting information might be a clearer approach? WDYT?

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,269 @@
...
+
+1. [Configuring Pipelines/Tekton with the configuration script after using our `ai-rhdh-installer`](#ai-rhdh-installer-script-configuration)
+2. [Configuring Pipelines/Tekton manually after using our `ai-rhdh-installer`](#ai-rhdh-installer-manual-configuration)
+3. [Configuring Pipelines/Tekton with the configuration script for a pre-existing Red Hat Developer Hub instance](#pre-existing-instance-script-configuration)
+4. [Configuring Pipelines/Tekton manually for a pre-existing Red Hat Developer Hub instance](#pre-existing-instance-manual-configuration)
+
+### Prerequisites
+
+- [yq](https://github.com/mikefarah/yq/) version 4.0+.
+- [kubectl](https://github.com/kubernetes/kubectl) version compatible with your target cluster
```
thepetk: I think `kubectl` should also be included in the `GITOPS-CONFIG.md` pre-requisites section. Might be better to mention that having `oc` installed is also ok?

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,269 @@
...
+
+### AI-RHDH-Installer: Script Configuration
+If you installed using the [`ai-rhdh-installer`](../README.md#install) all that is required for hooking up the Pipelines services and Tekton to your RHDH instance is to run `bash ./scripts/configure-tekton.sh`. Please note if you changed the installation namespace used by the installer you will first need to run `export NAMESPACE=<namespace used>` as the default value is `ai-rhdh`.
+
+### AI-RHDH-Installer: Manual Configuration
+
+#### Step 1: TektonConfig CR
+You should patch the `TektonConfig` CR called `config` with the extended configuration under [`tekton-config.json`](../resources/tekton-config.json)
```
thepetk, line 21: might be good to include an example command here?

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,269 @@
...
+            "transparency.url": "http://rekor-server.<rhdh_namespace>.svc"
+        },
+        ...
+    }
+}
+```
+
+#### Step 2: RHDH kubernetes plugin service account
+
+As part of the `ai-rhdh-installer` a service account with a token secret was created in your desired namespace with the name `rhdh-kubernetes-plugin`, token secret should have a name pattern `rhdh-kubernetes-plugin-token-*`, keep note of this Secret.
```
thepetk, line 45: Again the `ai-rhdh-installer` seems to be a pre-requisite might be better if we have it as required for both manual and script cases.

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### Step 3: Setting up deployment namespaces
+
+You will need to create the namespaces for the different app deployments. With a target namespace called `ai-rhdh`, these namespaces would be called `ai-rhdh-app-development`, `ai-rhdh-app-stage`, and `ai-rhdh-app-prod`. These namespaces can be created using the following:
```
thepetk: what if I use a different namespace other than `ai-rhdh`?

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,269 @@
...
+
+```sh
+kubectl -n $APP_NAMESPACE create secret generic "gitlab-auth-secret" \
+    --from-literal=password=<gitlab_pat> \
+    --from-literal=username=oauth2 \
+    --type=kubernetes.io/basic-auth \
+    --dry-run=client -o yaml | kubectl -n $APP_NAMESPACE apply --filename - --overwrite=true >/dev/null
+```
+
+Otherwise, you will need to create a Git authentication secret with your PAT (e.g. GitHub PAT) as follows:
```
thepetk, line 104: If I'm not wrong the PAT you'll need to apply here should have a specific list of permissions for the github org right? Or it's not the case here?

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+#### Step 7: Setting up quay image registry secret under deployment namespaces
+
+For accessing the quay image registry, you'll need to create a secret
+to store the docker config json file:
+
+```sh
+kubectl -n $APP_NAMESPACE create secret docker-registry "rhdh-image-registry-token" \
+    --from-file=.dockerconfigjson="<dockerconfig_jsonfile>" --dry-run=client -o yaml | \
```
thepetk, line 131: A description of the `dockerconfig_jsonfile` might be helpful for the user.

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,269 @@
...
+
+To include the [Tekton plugins list](../dynamic-plugins/tekton-plugins.yaml) we need to edit the dynamic plugins ConfigMap that was created by the RHDH Operator:
+
+![Dynamic Plugins Example](../assets/dynamic-plugins-example.png)
+
+Edit the associated `yaml` file to include the contents of the [Tekton plugins list](../dynamic-plugins/tekton-plugins.yaml) under the `plugins` section:
+
+![Dynamic Plugins Example 3](../assets/dynamic-plugins-example-3.png)
+
+Alternatively, we can use this series of commands to perform the same task with `kubectl` and `yq` using the [`tekton-plugins.yaml`](../dynamic-plugins/tekton-plugins.yaml):
```
thepetk, line 158: I think having sub-section might help here too

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,269 @@
...
+secrets:
+- name: rhdh-image-registry-token
+imagePullSecrets:
+- name: rhdh-image-registry-token
+" >/dev/null
+```
+
+#### Step 8: Updating Plugins
+
+To include the [Tekton plugins list](../dynamic-plugins/tekton-plugins.yaml) we need to edit the dynamic plugins ConfigMap that was created by the RHDH Operator:
```
thepetk, line 148: Out of curiosity, is there any difference if someone has installed RHDH through helm?

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,269 @@
...
+3. [Configuring Pipelines/Tekton with the configuration script for a pre-existing Red Hat Developer Hub instance](#pre-existing-instance-script-configuration)
+4. [Configuring Pipelines/Tekton manually for a pre-existing Red Hat Developer Hub instance](#pre-existing-instance-manual-configuration)
+
+### Prerequisites
+
+- [yq](https://github.com/mikefarah/yq/) version 4.0+.
+- [kubectl](https://github.com/kubernetes/kubectl) version compatible with your target cluster
+
+### AI-RHDH-Installer: Script Configuration
+If you installed using the [`ai-rhdh-installer`](../README.md#install) all that is required for hooking up the Pipelines services and Tekton to your RHDH instance is to run `bash ./scripts/configure-tekton.sh`. Please note if you changed the installation namespace used by the installer you will first need to run `export NAMESPACE=<namespace used>` as the default value is `ai-rhdh`.
```
thepetk:
```suggestion
If you installed using the [`ai-rhdh-installer`](../README.md#install) all that is required for hooking up the Pipelines services and Tekton to your RHDH instance is to run `bash ./scripts/configure-pipelines.sh`. Please note if you changed the installation namespace used by the installer you will first need to run `export NAMESPACE=<namespace used>` as the default value is `ai-rhdh`.
```

thepetk (changes requested):
I've focused mostly on the docs, so apologies I've left some mnor comments :D

Tried to go through the installation too. When running the pipelines I get this error:

```
* Configuring Namespaces: ._...The Secret "rhdh-image-registry-token" is invalid: data[.dockerconfigjson]: Invalid value: "<secret contents redacted>": invalid character '}' looking for beginning of object key string
.......The Secret "rhdh-image-registry-token" is invalid: data[.dockerconfigjson]: Invalid value: "<secret contents redacted>": invalid character '}' looking for beginning of object key string
.......The Secret "rhdh-image-registry-token" is invalid: data[.dockerconfigjson]: Invalid value: "<secret contents redacted>": invalid character '}' looking for beginning of object key string
....OK
```
It might be my fault here trying, I'm trying to add the same dockeconfig json same as the rhtap installer.

I'd expect that the installation should exit. WDYT?

docs/PIPELINES-CONFIG.md
michael-valdron: > Out of curiosity, is there any difference if someone has installed RHDH through helm?

@thepetk The only difference is that they'll need to follow the 'Pre-Existing Instance' setup as installing through helm is separate from our installer and the environment variable `RHDH_PLUGINS` should be set to the different ConfigMap name for the dynamic plugins, normally is called `dynamic-plugins`.

docs/PIPELINES-CONFIG.md
michael-valdron: Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/143b26952fac4d4be9bb4d0942fab099b69d7dc5

docs/PIPELINES-CONFIG.md
michael-valdron: > If I'm not wrong the PAT you'll need to apply here should have a specific list of permissions for the github org right? Or it's not the case here?

@thepetk Specific permissions in general, not towards the organization though the account with the PAT should have elevated permissions to the organization, I think just write maybe.

michael-valdron: > > If I'm not wrong the PAT you'll need to apply here should have a specific list of permissions for the github org right? Or it's not the case here?
> 
> @thepetk Specific permissions in general, not towards the organization though the account with the PAT should have elevated permissions to the organization, I think just write maybe.

RHDH org app creds will be used to administrate the org whereas the PAT is just to allow GitOps to perform actions against the existing resources if I'm understanding the setup correctly.

scripts/configure-pipelines.sh
```diff
@@ -0,0 +1,318 @@
...
+            --from-literal=password=$GITLAB__TOKEN \
+            --from-literal=username=oauth2 \
+            --type=kubernetes.io/basic-auth \
+            --dry-run=client -o yaml | kubectl -n $APP_NAMESPACE apply --filename - --overwrite=true >/dev/null
+        echo -n "."
+    fi
+    SECRET_NAME="gitops-auth-secret"
+    if [ -n "$GITOPS__GIT_TOKEN" ]; then
+        kubectl -n $APP_NAMESPACE create secret generic "$SECRET_NAME" \
+            --from-literal=password=$GITOPS__GIT_TOKEN \
```
maysunfaisal, line 238: We will need to also provide

```
--from-literal=username="oauth2" \
```

It is because if we create secret without username, the pipelines we are using on the main branch will be broken. See [comment](https://github.com/redhat-ai-dev/ai-lab-template/pull/36#discussion_r1767504108)

maysunfaisal: See the gitlab-auth-secret above for another reference where we provided username

docs/GITOPS-CONFIG.md
michael-valdron: @thepetk Yeah I think this sounds better :+1:

scripts/configure-pipelines.sh
Jdubrick: In the current state of the pipeline definitions I have there is no `gitops-auth-secret` since I removed it when it was causing issues, so whatever we decide to do for this I can adapt the definitions too. That being said if we don't include it here _eventually_ the pipelines on the `main` branch won't be broken since the pipeline definition will change

Jdubrick (commented):
I tried to follow the setup but it doesn't seem to be triggering the pipeline runs when I import [this template](https://github.com/Jdubrick/ai-lab-template/blob/test-pipeline-definitions/all.yaml) that has the new definitions. Is it something I did wrong in the setup or if you ran it does it work for you @michael-valdron ? I am currently going through to verify everything is as it should be

docs/PIPELINES-CONFIG.md
michael-valdron: Oops :face_with_open_eyes_and_hand_over_mouth:

docs/GITOPS-CONFIG.md
michael-valdron: > Here I think it gets confusing. Installing with `ai-rhdh-installer` is mentioned as pre-requisite for the script configuration. Reading this sentence makes me think it should be for both right?

@thepetk I'm not sure I fully understand the question, `ai-rhdh-installer` would be a prerequisite for all configure scripts as the configure scripts are ran after the fact, the only exception this is not the case is if the user already brought their own components in which case they need to follow the 'Pre-Existing Instance' steps. Does this touch on what you're asking?

Jdubrick: Yeah I believe there might have been some confusion, in this case it is the manual configuration for tekton if you installed the resources using the `ai-rhdh-installer` so it would be a prereq, like @michael-valdron said the `pre-existing instances` would have other options available!

docs/GITOPS-CONFIG.md
michael-valdron: Also done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/143b26952fac4d4be9bb4d0942fab099b69d7dc5

scripts/configure-pipelines.sh
michael-valdron: > In the current state of the pipeline definitions I have there is no `gitops-auth-secret` since I removed it when it was causing issues, so whatever we decide to do for this I can adapt the definitions too. That being said if we don't include it here _eventually_ the pipelines on the `main` branch won't be broken since the pipeline definition will change

@maysunfaisal @Jdubrick I wonder if this could be at least somewhat solved by adding an extra optional env var to provide a username and if its blank set up the classical way and if its not blank set up using the username flags. What do you all think?

michael-valdron: > > In the current state of the pipeline definitions I have there is no `gitops-auth-secret` since I removed it when it was causing issues, so whatever we decide to do for this I can adapt the definitions too. That being said if we don't include it here _eventually_ the pipelines on the `main` branch won't be broken since the pipeline definition will change
> 
> @maysunfaisal @Jdubrick I wonder if this could be at least somewhat solved by adding an extra optional env var to provide a username and if its blank set up the classical way and if its not blank set up using the username flags. What do you all think?

If there's more work to be done, we could look into separating into its own task to see what else is needed to get username based auth to work as well without removing classic compatibility. Only obstacle I still see with the toggle under the configure script is that once a deployment is configured either way it will limit it to just the software templates with whichever setup it supports.

docs/GITOPS-CONFIG.md
michael-valdron: Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/5659a91ba4bc3ba06bb2b68130e890d0d1fa24e9

docs/GITOPS-CONFIG.md
michael-valdron: > Might be better also here to be more detailed:
> 
>     * `If this is the case you can simply add that field and the necessary secret information`
> 
> 
> I think maybe add this info as troubleshooting information might be a clearer approach? WDYT?

@thepetk Yeah I agree, I think specifying `envFrom` again here would make it more clear, since there is a lot in this paragraph it made me think which field it was referring to as well. Also rewording to 'secret reference' rather than 'secret information'.

michael-valdron: Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/b2898c501fd7b26a578ec9cdd15f7da7c58655bf

michael-valdron:
> I tried to follow the setup but it doesn't seem to be triggering the pipeline runs when I import [this template](https://github.com/Jdubrick/ai-lab-template/blob/test-pipeline-definitions/all.yaml) that has the new definitions. Is it something I did wrong in the setup or if you ran it does it work for you @michael-valdron ? I am currently going through to verify everything is as it should be

@Jdubrick I can confirm the pipeline runs work with the current pipeline definition we're using from RHTAP (before your changes go in), the pipeline runs can sometimes be delayed I find for ~20 minutes before they trigger, I will try using your pipeline definitions myself as well to see what I can find.

Another thing to keep in mind is that if there are any errors during the initial flow of the component creation, it could be related to the fact we don't have RHDH configuration yet which could effect some of the steps of the ST setup.

docs/PIPELINES-CONFIG.md
michael-valdron: See comment: https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7#discussion_r1775403719

docs/PIPELINES-CONFIG.md
thepetk: > I think just write maybe.

Might be good to provide little context about the creds in case someone is trying to install and their token doesn't have the necessary perms?

docs/PIPELINES-CONFIG.md
gabemontero: the "non-trivial" TektonConfig used by Konflux is at https://github.com/redhat-appstudio/infra-deployments/blob/c06c88f16fbd5998ad9465ae0cc2612f0ffee248/components/pipeline-service/development/main-pipeline-service-configuration.yaml#L1631-L1840 fwiw

resources/tekton-config.json
```diff
@@ -0,0 +1,29 @@
...
+            "artifacts.pipelinerun.storage": "oci",
+            "artifacts.taskrun.format": "in-toto",
+            "artifacts.taskrun.storage": "oci",
+            "transparency.enabled": "true"
+        },
+        "platforms": {
+            "openshift": {
+                "pipelinesAsCode": {
+                    "settings": {
+                        "remember-ok-to-test": "false"
```
gabemontero: fwiw while RHTAP went with this, in part because of some back and forth wrt Konflux, this feature was not yet enabled in Konflux (I'll avoid the tl;dr and explain in the next parking lot if you like), as the Pipelines-as-code SME in openshift-pipelines / tekton did not necessarily agree with the approach

All that said, I'm not suggesting breaking away from what RHTAP is doing.

But there is some interesting devil in the details wrt this.

Lastly, cross reference with my other comment, the current Konflux Tekton Config:  https://github.com/redhat-appstudio/infra-deployments/blob/c06c88f16fbd5998ad9465ae0cc2612f0ffee248/components/pipeline-service/development/main-pipeline-service-configuration.yaml#L1631-L1840

docs/GITOPS-CONFIG.md
thepetk: @michael-valdron @Jdubrick sharing some more context as my comment might be missleading. Basically, what I mean is:

* [### AI-RHDH-Installer: Script Configuration](https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/b2898c501fd7b26a578ec9cdd15f7da7c58655bf/docs/GITOPS-CONFIG.md?plain=1#L17-L18) we explicitly mention that the user has to use `ai-rhdh-installer`.

* Tho in [## AI-RHDH-Installer: Manual Configuration](https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/b2898c501fd7b26a578ec9cdd15f7da7c58655bf/docs/GITOPS-CONFIG.md?plain=1#L20-L25) things are more implicit (at least this is my impression from my first pass of the guide).

I think it might be more clear if we mention in the beggining of both ways (manual, script) that the `ai-rhdh-installer` is a requirement, like we do in the [list of contents](https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/b2898c501fd7b26a578ec9cdd15f7da7c58655bf/docs/GITOPS-CONFIG.md?plain=1#L4-L9). WDYT?

docs/PIPELINES-CONFIG.md
thepetk: Resolving this one as I think the comments are duplicates :)

scripts/configure-pipelines.sh
Jdubrick: the username could just be hardcoded like it is in rhtap also @michael-valdron

chart/templates/developer-hub/subscription.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-{{ if (index .Values "developer-hub") }}
+{{ if and (index .Values "developer-hub") (index .Values "developer-hub" "enabled") }}
```
Jdubrick: Question: Is this the best way to write the conditional? It seems hard to follow at first glance but if that is how Helm suggests doing it then it's okay

chart/templates/openshift-gitops/subscription.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-{{ if (index .Values "openshift-gitops") }}
+{{ if and (index .Values "openshift-gitops") (index .Values "openshift-gitops" "enabled") }}
```
Jdubrick: Same as other question about the structure of the conditional

chart/templates/openshift-pipelines/subscription.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-{{ if (index .Values "openshift-pipelines") }}
+{{ if and (index .Values "openshift-pipelines") (index .Values "openshift-pipelines" "enabled") }}
```
Jdubrick: Same as other question

default-private.env
```diff
@@ -0,0 +1,7 @@
+# Private env vars
```
Jdubrick, line 1: If users want to use this file can they copy paste the value as normal? I know with multiline items such as the github private key it may be wonky?

docs/GITOPS-CONFIG.md
Jdubrick: In the wording of the content I suppose you're right but the sections are prefixed with the method, **`AI-RHDH-Installer`**, no?

docs/GITOPS-CONFIG.md
```diff
@@ -20,31 +20,45 @@ If you installed using the [`ai-rhdh-installer`](../README.md#install) all that
...
 #### Step 1: ConfigMaps and Secrets
-In your chosen namespace you should apply the following ConfigMaps:
-1.  [`argocd-config.yaml` ](../resources/argocd-config.yaml)
-2.  [`argocd-plugins.yaml`](../resources/argocd-plugins.yaml)
+In your chosen namespace you should apply the `argocd-config` ConfigMap under [`argocd-config.yaml` ](../resources/argocd-config.yaml).
 
 As part of the `ai-rhdh-installer` a secret was created in your desired namespace with the name `rhdh-argocd-secret`, keep note of this Secret as well as the two ConfigMaps applied above.
 
-#### Step 2: Updating Plugins
+#### Step 2.1: Updating Plugins with the web console
```
Jdubrick:
```suggestion
#### Step 2.1: Updating Plugins Via Web Console
```

docs/GITOPS-CONFIG.md
```diff
@@ -20,31 +20,45 @@ If you installed using the [`ai-rhdh-installer`](../README.md#install) all that
...
-Edit the associated `yaml` file to include `argocd-plugins.yaml` under the `includes` section:
+Edit the associated `yaml` file to include the contents of the [ArgoCD plugins list](../dynamic-plugins/argocd-plugins.yaml) under the `plugins` section:
 
 ![Dynamic Plugins Example 2](../assets/dynamic-plugins-example-2.png)
 
-#### Step 3: Updating RHDH Deployment
-Now that all of the required ConfigMaps and Secrets are apart of the namespace, as well as adding the plugins we must edit the RHDH Deployment to properly reference these items.
-
-Since we are referencing `argocd-plugins.yaml` in our dynamic-plugins ConfigMap we first must add `argocd-plugins.yaml` as a `volumeMount` to the `initContainers` field in the RHDH Deployment under `.spec.template.spec.initContainers.volumeMounts`
+#### Step 2.2: Updating Plugins with the CLI
```
Jdubrick:
```suggestion
#### Step 2.2: Updating Plugins Via CLI
```

docs/GITOPS-CONFIG.md
Jdubrick: Yes I believe even in the ai-rhdh-installer case you can change the namespace you deploy the chart too so for clarity we should probably make this `<namespace>` instead of a hard `ai-rhdh`

docs/PIPELINES-CONFIG.md
Jdubrick: If we add this we should add the requirement to gitops docs as well if it isn't already there

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,271 @@
...
+        "chain": {
+            ...
+            "transparency.url": "http://rekor-server.<rhdh_namespace>.svc"
+        },
+        ...
+    }
+}
+```
+
+#### Step 2: RHDH kubernetes plugin service account
```
Jdubrick:
```suggestion
#### Step 2: RHDH Kubernetes Plugin Service Account
```

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,271 @@
...
+
+As part of the `ai-rhdh-installer` a service account with a token secret was created in your desired namespace with the name `rhdh-kubernetes-plugin`, token secret should have a name pattern `rhdh-kubernetes-plugin-token-*`, keep note of this Secret.
+
+#### Step 3: Setting up deployment namespaces
```
Jdubrick:
```suggestion
#### Step 3: Setting Up Deployment Namespaces
```

docs/PIPELINES-CONFIG.md
Jdubrick: I believe if at the beginning you overwrote `$NAMESPACE` it is going to create `$NAMESPACE-development`, `$NAMESPACE-stage` and `$NAMESPACE-prod`

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,271 @@
...
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    argocd.argoproj.io/managed-by: $NAMESPACE
+  name: $APP_NAMESPACE
+EOF
+```
+
+#### Step 4: Setting up cosign secret under deployment namespaces
```
Jdubrick:
```suggestion
#### Step 4: Setting Up Cosign Secret Under Deployment Namespaces
```

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,271 @@
...
+        app.kubernetes.io/instance: default
+        app.kubernetes.io/part-of: tekton-chains
+        operator.tekton.dev/operand-name: tektoncd-chains
+    name: cosign-pub
+    namespace: $APP_NAMESPACE
+type: Opaque
+EOF
+```
+
+#### Step 5: Setting up gitops authentication secret under deployment namespaces
```
Jdubrick:
```suggestion
#### Step 5: Setting Up Gitops Authentication Secret Under Deployment Namespaces
```

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,271 @@
...
+Otherwise, you will need to create a Git authentication secret with your PAT (e.g. GitHub PAT) as follows:
+
+```sh
+kubectl -n $APP_NAMESPACE create secret generic "gitops-auth-secret" \
+    --from-literal=password=<git_pat> \
+    --type=kubernetes.io/basic-auth \
+    --dry-run=client -o yaml | kubectl -n $APP_NAMESPACE apply --filename - --overwrite=true >/dev/null
+```
+
+#### Step 6: Setting up pipelines secret under deployment namespaces
```
Jdubrick:
```suggestion
#### Step 6: Setting Up Pipelines Secret Under Deployment Namespaces
```

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,271 @@
...
+
+```sh
+kubectl -n $APP_NAMESPACE create secret generic "pipelines-secret" \
+    --from-literal=webhook.secret=<webhook_secret> \
+    --dry-run=client -o yaml | kubectl -n $APP_NAMESPACE apply --filename - --overwrite=true >/dev/null
+```
+
+#### Step 7: Setting up quay image registry secret under deployment namespaces
+
+For accessing the quay image registry, you'll need to create a secret
```
Jdubrick: Nit: this sentence shows broken up on 2 different lines

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,271 @@
...
+
+You will need to create a pipeline secret containing the webhook secret for the Git organization for deployments to have proper access via the tekton pipelines:
+
+```sh
+kubectl -n $APP_NAMESPACE create secret generic "pipelines-secret" \
+    --from-literal=webhook.secret=<webhook_secret> \
+    --dry-run=client -o yaml | kubectl -n $APP_NAMESPACE apply --filename - --overwrite=true >/dev/null
+```
+
+#### Step 7: Setting up quay image registry secret under deployment namespaces
```
Jdubrick:
```suggestion
#### Step 7: Setting Up Quay Image Registry Secret Under Deployment Namespaces
```

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,271 @@
...
+```sh
+kubectl -n $APP_NAMESPACE patch serviceaccounts "<default|pipeline>" --patch "
+secrets:
+- name: rhdh-image-registry-token
+imagePullSecrets:
+- name: rhdh-image-registry-token
+" >/dev/null
+```
+
+#### Step 8.1: Updating Plugins with the web console
```
Jdubrick:
```suggestion
#### Step 8.1: Updating Plugins Via Web Console
```

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,271 @@
...
+
+To include the [Tekton plugins list](../dynamic-plugins/tekton-plugins.yaml) we need to edit the dynamic plugins ConfigMap that was created by the RHDH Operator:
+
+![Dynamic Plugins Example](../assets/dynamic-plugins-example.png)
+
+Edit the associated `yaml` file to include the contents of the [Tekton plugins list](../dynamic-plugins/tekton-plugins.yaml) under the `plugins` section:
+
+![Dynamic Plugins Example 3](../assets/dynamic-plugins-example-3.png)
+
+#### Step 8.2: Updating Plugins with the CLI
```
Jdubrick:
```suggestion
#### Step 8.2: Updating Plugins Via CLI
```

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,271 @@
...
+5. If the secret is not marked as immutable, make it so
+    ```sh
+    kubectl patch secret -n "openshift-pipelines" "signing-secrets" \
+        --dry-run=client -o yaml \
+        --patch='{"immutable": true}' \
+        | kubectl apply -f - >/dev/null
+    ````
+
+
+#### Step 4: Run configure script
```
Jdubrick:
```suggestion
#### Step 4: Run Configure Script
```

dynamic-plugins/tekton-plugins.yaml
```diff
@@ -0,0 +1,45 @@
+plugins:
```
Jdubrick, line 1: Question: the remaining dynamic plugins will probably be installed with the rhdh config script? such as topology

scripts/configure-pipelines.sh
```diff
@@ -0,0 +1,318 @@
...
+until [ ! -z "${GITHUB__APP__PRIVATE_KEY}" ]; do
+    read -p "Enter your GitHub App Private Key (Use CTRL-D when finished): " -d $'\04' GITHUB__APP__PRIVATE_KEY
+    if [ -z "${GITHUB__APP__PRIVATE_KEY}" ]; then
+        echo "No GitHub App Private Key entered, try again."
+    fi
+done
+
+# Reads Git PAT
+# Optional: If left blank during user prompt, the namespace secret will not be created
+if [ -z "${GITOPS__GIT_TOKEN}" ]; then
```
Jdubrick, line 71: Nit: When doing the script and manually inputting the values via CLI the end of the private key and the gitops prompt are on the same line:

```
-----END RSA PRIVATE KEY-----^DEnter your Git Token (Optional): <redacted token>
```

Jdubrick (changes requested):
I left a few comments but all of it is pretty minor stuff. I also was able to get everything deployed and can confirm the pipelines are running with your setup and my definitions (see below), I used the installer with my own RHDH setup. I tested the initial trigger when creating a component as well as when I opened a PR it triggered and when the PR was merged it properly created a new image and deployed it, working alongside Argo!

Screenshot 2024-09-25 at 3 05 38 PM

chart/templates/developer-hub/subscription.yaml
thepetk: That's interesting, cause according to [helm documentation](https://docs.helm.sh/docs/chart_template_guide/control_structures/#ifelse) it should return `false` for empty or null values. Not sure what happens if the parent value doesn't exist.

scripts/configure-pipelines.sh
michael-valdron: Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/6d03b6bd7ba3a8f9aaf80253a88ed23d2518dee8

dynamic-plugins/tekton-plugins.yaml
michael-valdron: @Jdubrick Yes, I'll create a `developer-hub-plugins.yaml` or `plugins.yaml` in the changes for the next issue.

docs/PIPELINES-CONFIG.md
michael-valdron: > I believe if at the beginning you overwrote `$NAMESPACE` it is going to create `$NAMESPACE-development`, `$NAMESPACE-stage` and `$NAMESPACE-prod`

@Jdubrick @thepetk Missing the 'app' in between, it would be `$NAMESPACE-app-development`, `$NAMESPACE-app-stage` and `$NAMESPACE-app-prod`, `$NAMESPACE` is the target for the RHDH deployment which is being used as a prefix to the app deployment namespaces.

As suggested in another part I could just use the placeholder `<rhdh_namespace>` instead of the example `ai-rhdh`.

docs/PIPELINES-CONFIG.md
michael-valdron: > might be good to include an example command here?

Yeah I think I missed this part.

resources/tekton-config.json
michael-valdron: > fwiw while RHTAP went with this, in part because of some back and forth wrt Konflux, this feature was not yet enabled in Konflux (I'll avoid the tl;dr and explain in the next parking lot if you like), as the Pipelines-as-code SME in openshift-pipelines / tekton did not necessarily agree with the approach
> 
> All that said, I'm not suggesting breaking away from what RHTAP is doing.
> 
> But there is some interesting devil in the details wrt this.
> 
> Lastly, cross reference with my other comment, the current Konflux Tekton Config: https://github.com/redhat-appstudio/infra-deployments/blob/c06c88f16fbd5998ad9465ae0cc2612f0ffee248/components/pipeline-service/development/main-pipeline-service-configuration.yaml#L1631-L1840

@gabemontero I agree that I think its worth updating our tekton config to match the most recent one, though I wonder if this should be a separate task to avoid set backs in compatibility with our templates, wanna make sure it works as it did with the rhtap-installer. Though I am going to try to remove `openshift.pipelinesAsCode.settings.remember-ok-to-test`, as I suspect it won't matter from my understanding of what this does.

docs/PIPELINES-CONFIG.md
michael-valdron: Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/17c441450331fba50a78a680b3cfc238a4275a47

default-private.env
michael-valdron: @Jdubrick They'll need to ensure the value is surrounded by single quotes and if the value itself has single quotes in it these will need to be manually escaped for now. I think it would be good to create a helper script for properly setting the values just as rhtap-installer had, but I think this would be too out of scope for this PR.

docs/PIPELINES-CONFIG.md
michael-valdron: > > I think just write maybe.
> 
> Might be good to provide little context about the creds in case someone is trying to install and their token doesn't have the necessary perms?

@thepetk I'd prefer we link to a RHDH doc page which provide this information, e.g. 'More information about setting up the PAT token can be found <here|url.to.rhdh.docpage>', as the scope of these docs is to just provide instruction on how to use the installer and configure scripts.

docs/PIPELINES-CONFIG.md
michael-valdron: Fixed in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/8b46981fcee936554bd5dbb95cd294caa83734d8

chart/templates/developer-hub/subscription.yaml
michael-valdron: @Jdubrick @thepetk I've simplified the conditions to just checking for the child https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/d1ad0e77d9f9b9fbe95bbc1aa0b1a56d47f78fed, going to fully verify this a bit later. Still works when enabled.

docs/PIPELINES-CONFIG.md
michael-valdron: Fixed in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/b3f58b158927d574e6df3224dd13a160744a014e

docs/PIPELINES-CONFIG.md
thepetk: Yes that would be perfect!

docs/PIPELINES-CONFIG.md
michael-valdron: @thepetk How does this sound? https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/823d0247d4e2e44b14559b07de89f748c601ce43

Let me know if there is anything you'd like me to expand further on.

docs/PIPELINES-CONFIG.md
```diff
@@ -0,0 +1,275 @@
...
+  - Name of the dynamic plugins ConfigMap
+
+#### Step 3: Configure cosign
+
+The installer will configure and set up cosign that the `configure-pipeline.sh` script needs to reference the cosign public key to set to the cosign secret under the deployment namespaces.
+
+To configure cosign to sign secrets follow these steps:
+1. Download `cosign` cli tool
+    ```sh
+    curl -L https://github.com/sigstore/cosign/releases/latest/download/cosign-linux-amd64 -o cosign && chmod +x cosign
```
Jdubrick: I receive an exec format error on the binary, we may need to have this dynamically set depending which architecture the script is being run on?

chart/templates/developer-hub/subscription.yaml
michael-valdron: > @Jdubrick @thepetk I've simplified the conditions to just checking for the child [d1ad0e7](https://github.com/redhat-ai-dev/ai-rhdh-installer/commit/d1ad0e77d9f9b9fbe95bbc1aa0b1a56d47f78fed), going to fully verify this a bit later. Still works when enabled.

@Jdubrick @thepetk Had to backtrack this as it errors if parent is not set, so in the first condition I changed from `index` to `hasKey` to only check if the parent key exists rather than trying to get the value: https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/b54296d7fafed04b44cb1357418bf671094c6bc6

As per what @thepetk said I think returning null on non-existence (treated as `false`) on the second statement makes sense, this is the simplest alternative method I could find with helm conditions like this.

chart/templates/openshift-gitops/subscription.yaml
michael-valdron: See comment https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7#discussion_r1779167022

chart/templates/openshift-pipelines/subscription.yaml
michael-valdron: See comment https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7#discussion_r1779167022

docs/PIPELINES-CONFIG.md
michael-valdron: Revised in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/48b5d6824d29fd553c53e384f767461c58a78a94

docs/GITOPS-CONFIG.md
michael-valdron: @Jdubrick @thepetk I decided to use the env var `$NAMESPACE` that refers to the rhdh namespace https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/b8753424110732cc94a6f2aec5575f72240e2392 as this step of setting it is highlighted at the beginning of each of these documents:
- [PIPELINES-CONFIG.md](https://github.com/michael-valdron/ai-rhdh-installer/blob/b8753424110732cc94a6f2aec5575f72240e2392/docs/PIPELINES-CONFIG.md?plain=1#L16)
- [GITOPS-CONFIG.md](https://github.com/michael-valdron/ai-rhdh-installer/blob/b8753424110732cc94a6f2aec5575f72240e2392/docs/GITOPS-CONFIG.md?plain=1#L19)

thepetk: sgtm

docs/PIPELINES-CONFIG.md
michael-valdron: Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/3fce036f294606cfefaddfc6a05ed534b235fdd3

docs/PIPELINES-CONFIG.md
```diff
@@ -210,9 +210,9 @@ You are able to store these values in environment variables or paste them during
 The installer will configure and set up cosign that the `configure-pipeline.sh` script needs to reference the cosign public key to set to the cosign secret under the deployment namespaces.
 
 To configure cosign to sign secrets follow these steps:
-1. Download `cosign` cli tool
+1. Set the architecture you wish to use `ARCH=<architecture>`, for example `amd64` or `arm64`, and download `cosign` cli tool
     ```sh
-    curl -L https://github.com/sigstore/cosign/releases/latest/download/cosign-linux-amd64 -o cosign && chmod +x cosign
+    curl -L https://github.com/sigstore/cosign/releases/latest/download/cosign-linux-$ARCH -o cosign && chmod +x cosign
```
Jdubrick: You beat me to it before I added my review but we will need to do one step further, if you are on Mac you need the `darwin` version instead of `linux`

`cosign-darwin-arm64` for example

Jdubrick (changes requested):

Jdubrick:
Screenshot 2024-09-30 at 10 59 10 AM

Pipelines running when following the pre-existing instructions as well, looks like it is good for both scenarios on the pipeline run end! Should be good after the final comments are addressed on my end :)

docs/PIPELINES-CONFIG.md
michael-valdron: Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/01797088754c61e568ae65278b580eb433f9b6a6

Jdubrick (approved):
lgtm. Since this was such a large PR and lots of people had comments I would say we get approval from the others as well.

thepetk:
> lgtm. Since this was such a large PR and lots of people had comments I would say we get approval from the others as well.

@michael-valdron @Jdubrick I agree I can take a look tomorrow and try to set everything up and test again the PR (after the updates made during review)

resources/tekton-config.json
michael-valdron: > > fwiw while RHTAP went with this, in part because of some back and forth wrt Konflux, this feature was not yet enabled in Konflux (I'll avoid the tl;dr and explain in the next parking lot if you like), as the Pipelines-as-code SME in openshift-pipelines / tekton did not necessarily agree with the approach
> > All that said, I'm not suggesting breaking away from what RHTAP is doing.
> > But there is some interesting devil in the details wrt this.
> > Lastly, cross reference with my other comment, the current Konflux Tekton Config: https://github.com/redhat-appstudio/infra-deployments/blob/c06c88f16fbd5998ad9465ae0cc2612f0ffee248/components/pipeline-service/development/main-pipeline-service-configuration.yaml#L1631-L1840
> 
> @gabemontero I agree that I think its worth updating our tekton config to match the most recent one, though I wonder if this should be a separate task to avoid set backs in compatibility with our templates, wanna make sure it works as it did with the rhtap-installer. Though I am going to try to remove `openshift.pipelinesAsCode.settings.remember-ok-to-test`, as I suspect it won't matter from my understanding of what this does.

Tested removing `openshift.pipelinesAsCode.settings.remember-ok-to-test` and found behaviour is the same, opted to remove this to keep consistency: https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/999983a15fcc8751cae5314b4fbc9357f4861355

docs/PIPELINES-CONFIG.md
michael-valdron: Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/fa8bec73487eb86784456959ff288f94ff59a927

resources/tekton-config.json
gabemontero: Interesting @michael-valdron .... perhaps an openshift pipelines bug :-).

A minor follow up suggestion: I'm fine if you do not want to pursue, but if
- you converted this json file to its yaml alternative, so you can then put comments (https://stackoverflow.com/questions/244777/can-comments-be-used-in-json)
- and then you could put a comment explaining your findings here, so we have some historical reference "in the code" that explains things if a question in this space comes up in the future


WDYT?

thanks

maysunfaisal:
@michael-valdron Please squash merge when merging this PR

thepetk (approved):
LGTM

After some effort due to a strange issue with RHDH I was able to confirm that the PR works.

More detailed:

* I've created an RHDH instance by using the `ai-rhdh-installer`.
* Added github authentication to the instance by configuring the `app-config.yaml`.
* Ran both scripts for gitops & pipelines.
* **Important step**: The template should mention as default namespace `ai-rhdh` ([L#277](https://github.com/thepetk/ai-lab-template/blob/temp_chatbot/templates/chatbot-temp/template.yaml#L277), [L#287](https://github.com/thepetk/ai-lab-template/blob/temp_chatbot/templates/chatbot-temp/template.yaml#L287) & [L#373](https://github.com/thepetk/ai-lab-template/blob/temp_chatbot/templates/chatbot-temp/template.yaml#L373)).

* All steps passed successfully
* ArgoCD worked fine:
Screenshot from 2024-10-01 15-56-03

* GitOps worked fine:
Screenshot from 2024-10-01 15-56-33

resources/tekton-config.json
michael-valdron: @gabemontero  Unfortunately, `kubectl` does not recognize the `_comment` property as a comment property to ignore:
```
Error from server (BadRequest): admission webhook "webhook.operator.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "_comment"
FAIL
```

To provide comments I've decided to convert this into a yaml file that gets processed into json with `yq` (already does this) to be passed into the `kubectl patch` command to patch the TektonConfig CR: https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/f626d13969115ab60298755cd7fb1f804ec78ae3

With this, I've provided the comment you requested as a reference point in case we want to revert for any reason.

README.md
```diff
@@ -43,4 +43,8 @@ This installer is incompatible with `default` namespace installations, install a
 
 ## GitOps/ArgoCD Configuration
 
-Detailed documentation for configuring GitOps/ArgoCD can be found in [`GITOPS-CONFIG.md`](./docs/GITOPS-CONFIG.md)
\ No newline at end of file
+Detailed documentation for configuring GitOps/ArgoCD can be found in [`GITOPS-CONFIG.md`](./docs/GITOPS-CONFIG.md)
```
maysunfaisal, line 46: Should we provide a bit more info on setting up ENVs? Glancing at the README, I could not find any mention of how to set up https://github.com/michael-valdron/ai-rhdh-installer/blob/feat/pipelines-config/default-private.env 🤔

gabemontero (approved):
/lgtm

README.md
maysunfaisal: If this is TODO, please feel free to merge PR with other LGTMs received

michael-valdron: Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/7/commits/21c11e362b2e1fd47ee59025c77d1360c7cee3e7

maysunfaisal (approved):

PR #8, opened 2024-10-04
DEVAI-31: Custom Namespace Tekton Task

michael-valdron:
### What does this PR do?:

Adds a Tekton Task to the RHDH namespace during run of `configure-pipelines.sh` so that users can target different namespaces for their apps and the Tekton Task will setup up that target namespace with the resources it needs to manage the deployments.

resolves missing part of https://issues.redhat.com/browse/DEVAI-31

### PR acceptance criteria:

- [x] Tested and Verified

  

### DIFF
--- a/docs/PIPELINES-CONFIG.md
+++ b/docs/PIPELINES-CONFIG.md
@@ -44,9 +44,169 @@ cat resources/tekton-config.json | kubectl patch tektonconfig config --type 'mer
 
 As part of the `ai-rhdh-installer` a service account with a token secret was created in your desired namespace with the name `rhdh-kubernetes-plugin`, token secret should have a name pattern `rhdh-kubernetes-plugin-token-*`, keep note of this Secret.
 
-#### Step 3: Setting Up Deployment Namespaces
+#### Step 3: Create App Namespace Setup Task
 
-You will need to create the namespaces for the different app deployments. You will need a namespace for each kind of the deployment: development, staging, and production. These namespaces would follow a naming pattern of `<rhdh_namepsace>-app-<deployment_kind>` with the target RHDH namespace as the prefix, therefore `$NAMESPACE-app-development`, `$NAMESPACE-app-stage`, and `$NAMESPACE-app-prod`. These namespaces can be created using the following:
+In order to set up target app namespaces for the software template created components, you will need to create a Tekton Task that will trigger when the app is created from the template. You can use [dev-setup-task.yaml](../resources/dev-setup-task.yaml) as the starting point.
+
+First you will need to set the default values under `.spec.params`, second you will need to fetch the cosign signing public key, third set `.spec.steps[0].script` to a script that will create the needed secret resources when the Task is run.
+
+##### Step 3.1: Set Git Token Default Value
+
+The `git_token` parameter is set to a Personal Access Token \(PAT\) that is tied to an account that accesses the RHDH Git repositories, such as a GitHub account PAT. The default value can be set by setting the `default` field:
+
+```yaml
+- default: '<git_pat>'
+  description: |
+    Git token
+  name: git_token
+  type: string
+```
+
+##### Step 3.2: Set GitLab Token Default Value
+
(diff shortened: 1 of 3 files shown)


thepetk:
Cool feature!

docs/PIPELINES-CONFIG.md
```diff
@@ -44,9 +44,169 @@ cat resources/tekton-config.json | kubectl patch tektonconfig config --type 'mer
...
+```
+
+See the following optional steps for more information about each secret being created in this Task:
+1. [Setting Up Gitops Authentication Secret Under Deployment Namespaces](#step-6-setting-up-gitops-authentication-secret-under-deployment-namespaces-optional)
+2. [Setting Up Pipelines Secret Under Deployment Namespaces](#step-7-setting-up-pipelines-secret-under-deployment-namespaces-optional)
+3. [Setting Up Quay Image Registry Secret Under Deployment Namespaces](#step-8-setting-up-quay-image-registry-secret-under-deployment-namespaces-optional)
+
+#### Step 4: Setting Up Deployment Namespaces \(Optional\)
+
+Optionally, you can create the namespaces for the different app deployments, this will mirror some the steps for the Tekton Task created. You will need a namespace for each kind of the deployment: development, staging, and production. These namespaces would follow a naming pattern of `<rhdh_namepsace>-app-<deployment_kind>` with the target RHDH namespace as the prefix, therefore `$NAMESPACE-app-development`, `$NAMESPACE-app-stage`, and `$NAMESPACE-app-prod`. These namespaces can be created using the following:
```
Jdubrick, line 209: Nit/Question: Since we use the full word for `dev` -> `development` should we be using `stage` -> `staging` and `prod` -> `production`?

Jdubrick: Keep in mind if you opt to change this then there is multiple occurrences throughout this PR but I'd say it is optional to change

Jdubrick (approved):
Overall lgtm with a nit optional suggestion. I ran all of these with you already and here is the results!

All secrets created in new namespace:
Screenshot 2024-10-04 at 5 35 45 PM

Pipelines running fully in new namespace:
Screenshot 2024-10-04 at 5 41 09 PM

App deployed:
Screenshot 2024-10-04 at 5 42 50 PM

Namespace setup pipeline ran:
Screenshot 2024-10-04 at 5 43 27 PM

docs/PIPELINES-CONFIG.md
michael-valdron: @Jdubrick Yeah I would opt out of making this change this late since it would affect the technical part of how it works. Also, based on prior discussions this part is planned to be removed anyway.

resources/dev-setup-task.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: dev-namespace-setup
```
maysunfaisal, line 4: we need to coordinate the changes, the task reference here still mentions `rhtap-dev-namespace-setup` https://github.com/redhat-ai-dev/ai-lab-app/blob/main/templates/http/base/initialize-namespace.yaml#L36

Jdubrick: I can loop that change into https://issues.redhat.com/browse/DEVAI-168 as part of the switch from rhtap installer to our installer 

cc @maysunfaisal @michael-valdron

maysunfaisal (approved):
Thanks for the quick turnaround. 

Have you tried it out to see if it works?
Besides the task name reference, I could see the multi line script being added to the Task could be probelmatic due to new line and/or tabs.

resources/dev-setup-task.yaml
maysunfaisal: sure @Jdubrick

Jdubrick: I know the changes that have to be made since I made them to test this task out to produce my screenshots above anyway so it won't be any issue. If we change them now then the templates will break until the switch is made

michael-valdron:
> Besides the task name reference, I could see the multi line script being added to the Task could be probelmatic due to new line and/or tabs.

@maysunfaisal I don't think it would here as `yq` takes care of it.

PR #9, opened 2024-10-07
add script to install lightspeed plugin

jrichter1:
### What does this PR do?:
Adds a script to install/configure lightspeed dynamic plugin to RHDH. As part of this, the backstage config had to be modified to contain base URL for the app and the CORS settings to enable dynamic plugin asset loading.

### PR acceptance criteria:

- [x] Tested and Verified

  


Jdubrick:
I wonder if this should be baked into https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10 or if we rebase / revisit this PR after it is merged? It looks to me though like we could just install Lightspeed along with all the other plugins? wdyt @michael-valdron

michael-valdron (commented):
> I wonder if this should be baked into #10 or if we rebase / revisit this PR after it is merged? It looks to me though like we could just install Lightspeed along with all the other plugins? wdyt @michael-valdron

Yeah we should wait on #10 before we introduce lightspeed install as the changes from #10 will change the process a bit.

michael-valdron (commented):
@jrichter1 You'll need to rebase to the latest changes in `main`, I can provide feedback on this PR once this is done.

jrichter1:
@michael-valdron I've reworked the configuration into the new script, feel free to take a look :)

README.md
```diff
@@ -65,6 +65,10 @@ Detailed documentation for configuring Pipelines/Tekton can be found in [`PIPELI
 
 Detailed documentation for configuring Developer Hub can be found in [`RHDH-CONFIG.md`](./docs/RHDH-CONFIG.md)
 
+### Lightspeed Plugin Configuration
+
+Lightspeed plugin is installed as part of developer hub configuration script, if the `LIGHTSPEED_MODEL_URL` environment variable is set. Optionally, set the variable `LIGHTSPEED_API_TOKEN` for authenticated communication with the model service.
```
Jdubrick: nit

```suggestion
The Lightspeed plugin is installed as part of the Developer Hub configuration script if the `LIGHTSPEED_MODEL_URL` environment variable is set. Optionally, set the variable `LIGHTSPEED_API_TOKEN` for authenticated communication with the model service.
```

resources/lightspeed-plugins.yaml
```diff
@@ -0,0 +1,19 @@
+plugins:
```
Jdubrick: Our other plugins are kept in the `dynamic-plugins` dir, should we store this there too for consistency?

scripts/configure-dh.sh
```diff
@@ -131,6 +131,30 @@ fi
...
+fi
+if [ ! -z "${LIGHTSPEED_MODEL_URL}" ]; then
+    # make sure the target url ends in /v1
+    if ! [[ $LIGHTSPEED_MODEL_URL =~ \/v1$ ]]; then
+        LIGHTSPEED_MODEL_URL="$LIGHTSPEED_MODEL_URL/v1"
+    fi
+
+    plugin_sha=$(npm view @janus-idp/backstage-plugin-lightspeed dist.integrity)
+
+    cp -f $BASE_DIR/resources/lightspeed-plugins.yaml $BASE_DIR/dynamic-plugins/
```
Jdubrick: If we move the plugins to `/dynamic-plugins` do we need the cp line here?

scripts/configure-dh.sh
```diff
@@ -331,6 +368,7 @@ for f in $BASE_DIR/dynamic-plugins/*.yaml; do
 
     echo "OK"
 done
+rm -f $BASE_DIR/dynamic-plugins/lightspeed-plugins.yaml
```
Jdubrick: Same as comment re: keeping them in the `/dynamic-plugins` dir, if we move them we won't need this line I believe

Jdubrick (commented):
Overall lgtm, left some small comments / questions in the review. I will run through the setup today and see how it works!

resources/lightspeed-plugins.yaml
jrichter1: I was trying to make the installation optional, only if at least the model url is set. If I left the file in `dynamic-plugins`, the plugin would get installed regardless, even without the proxy setting.

I have no strong feelings about how it should be done, but I'd like to know your opinion on this.

Jdubrick: Ah yes that is a good point, at the end of the day it doesn't matter if we leave it in the `/resources` dir but maybe for clarity to separate plugins from other config files we could put it into a `/optional-plugins` dir?

wdyt @jrichter1 @michael-valdron , I am indifferent though and it doesn't _really_ matter

michael-valdron: > maybe for clarity to separate plugins from other config files we could put it into a `/optional-plugins` dir?

I'm liking this idea so we can still seperate the plugin lists from the kubernetes/openshift resource specs.

scripts/configure-dh.sh
michael-valdron: Instead of copying into `/dynamic-plugins` we could add a list of file paths to extend the base list from `/dynamic-plugins`, e.g. if `LIGHTSPEED_MODEL_URL` is defined, add `$BASE_DIR/optional-plugins/lightspeed-plugins.yaml` file to the extended list to iterate over.

scripts/configure-dh.sh
```diff
@@ -131,6 +131,30 @@ fi
...
 
+# Reads Model URL for lightspeed plugin
+# Optional: If no URL is entered, lightspeed plugin will not be configured
+if [ -z "${LIGHTSPEED_MODEL_URL}" ]; then
+    read -p "Enter your model URL for lightspeed (Optional): " LIGHTSPEED_MODEL_URL
+
+    # Reads API token for lightspeed plugin
+    # Optional: If no token is entered, lightspeed plugin will not use authenticated communication
+    if [ -z "${LIGHTSPEED_API_TOKEN}" ]; then
+        read -p "Enter API token for lightspeed (Optional): " LIGHTSPEED_API_TOKEN
```
michael-valdron: Nit: Would be good to add `LIGHTSPEED_MODEL_URL` and `LIGHTSPEED_API_TOKEN` to the secret variable definitions above.

scripts/configure-dh.sh
michael-valdron: This line could be removed if going with my suggestions. Otherwise if keeping the file copy approach this can be kept in.

michael-valdron (commented):
Great work on these changes! I'll also try to run these changes later today.

scripts/configure-dh.sh
jrichter1: moved to the `optional-plugins` approach, it does look better

Jdubrick (approved):
lgtm, I was able to easily get everything setup and have Lightspeed running for me within RHDH without a hitch.

@michael-valdron this can be merged once you have a chance to test it out as well and give your thoughts

scripts/configure-dh.sh
```diff
@@ -131,6 +133,28 @@ fi
...
+fi
+if [ ! -z "${LIGHTSPEED_MODEL_URL}" ]; then
+    # make sure the target url ends in /v1
+    if ! [[ $LIGHTSPEED_MODEL_URL =~ \/v1$ ]]; then
+        LIGHTSPEED_MODEL_URL="$LIGHTSPEED_MODEL_URL/v1"
+    fi
+
+    plugin_sha=$(npm view @janus-idp/backstage-plugin-lightspeed dist.integrity)
+    yq -i ".plugins.[0].integrity = \"${plugin_sha}\"" $BASE_DIR/optional-plugins/lightspeed-plugins.yaml
+fi
```
michael-valdron: I'm wondering just the same as we have for GitHub and GitLab integration if we should include an enablement variable so that we are not prompting the user for multiple optional inputs that they don't want enabled.

Something like:

```suggestion
if [[ ${LIGHTSPEED_INTEGRATION} == "true" ]]; then
    until [ ! -z "${LIGHTSPEED_MODEL_URL}" ]; do
        read -p "Enter your model URL for lightspeed: " LIGHTSPEED_MODEL_URL
        if [ -z "${LIGHTSPEED_MODEL_URL}" ]; then
            echo "No model URL for lightspeed entered, try again."
        fi
    done
    
    # Reads API token for lightspeed plugin
    # Optional: If no token is entered, lightspeed plugin will not use authenticated communication
    if [ -z "${LIGHTSPEED_API_TOKEN}" ]; then
        read -p "Enter API token for lightspeed (Optional): " LIGHTSPEED_API_TOKEN
    fi
fi

echo "OK"

if [[ ${LIGHTSPEED_INTEGRATION} == "true" ]]; then
    # make sure the target url ends in /v1
    if ! [[ $LIGHTSPEED_MODEL_URL =~ \/v1$ ]]; then
        LIGHTSPEED_MODEL_URL="$LIGHTSPEED_MODEL_URL/v1"
    fi

    plugin_sha=$(npm view @janus-idp/backstage-plugin-lightspeed dist.integrity)
    yq -i ".plugins.[0].integrity = \"${plugin_sha}\"" $BASE_DIR/optional-plugins/lightspeed-plugins.yaml
fi
```

michael-valdron: Should also put these prompts under `configure.sh` to prompt at the beginning of the run.

michael-valdron (commented):
Tried this myself as well, works good, I was able to reach the prompt with lightspeed and connected model service.

Only my last feedback about the input prompt and control logic before I sign off.

scripts/configure-dh.sh
jrichter1: Assuming the prompts would exist in both `configure.sh` and `configure-dh.sh`, any unfilled optional variable from the first prompt would fall through to the second script as unset again.

Some other value would be needed to discriminate between an optional value that wasn't processed, and one that was left empty on purpose. Otherwise we'll get duplicate prompts when calling `configure.sh`. Got any in mind?

jrichter1: Anyway, here is the enablement variable and prompt in `configure.sh`

michael-valdron: > any unfilled optional variable from the first prompt would fall through to the second script as unset again

@jrichter1 This is a good point, the optional input would prompt twice if left blank the first time. To counter this, it might make sense in each of the scripts under `/scripts` to include an additional `BYPASS_OPTIONAL_INPUT` flag variable to include in the checks for the optional prompts alongside the set conditions.

`configure.sh` can have this set going into each of these scripts to bypass the same optional input prompts regardless if they are set.

jrichter1: I've basically listed all the processed optional variables from `config.sh` into `BYPASS_OPTIONAL_INPUT`, so if any of them is found in there, its second prompt will be skipped. Can you give it a try?

Jdubrick: I was able to get everything up and running with the new additions, my lgtm still stands @jrichter1 @michael-valdron

michael-valdron (approved):
/lgtm

jrichter1:
@Jdubrick @michael-valdron any chance to get this merged?

README.md
```diff
@@ -102,6 +106,11 @@ Configuration scripts can either take user input or can have environment variabl
         - Toggle GitHub integration (login). Accepts `true` or `false`
     - `RHDH_GITLAB_INTEGRATION`
         - Toggle GitLab integration (login). Accepts `true` or `false`
+    - `LIGHTSPEED_MODEL_URL`
+        - Target model URL for lightspeed plugin
+    - `LIGHTSPEED_API_TOKEN`
+        - API token for lightspeed plugin model service
+3. Run `source private.env` to set all set environment variables within `private.env`
```
Jdubrick: I believe with new changes from the commit below we can remove this line:
https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/34

We should also include these env vars as part of the auto create for a private.env file for the user:
https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/main/scripts/store-private-env.sh#L5

Jdubrick: `LIGHTSPEED_INTEGRATION` would also do well in the list of env vars as well IMO, I believe in our manual testing we thought it'd be best to have all of the env vars that can be set able to be preset in a env file for example

Jdubrick (commented):
Just some minor comments since this repo has changed a bit over time!

README.md
jrichter1: done

Jdubrick (approved):
New changes lgtm, @michael-valdron you previously approved so after your re-approval we can merge this

michael-valdron (approved):
/lgtm

PR #10, opened 2024-10-07
DEVAI-38: Add Developer Hub Configuration Script

michael-valdron:
### What does this PR do?:

This PR allows a user to configure RHDH with the necessary config and plugins for AI Software Templates.

resolves https://issues.redhat.com/browse/DEVAI-38

### PR acceptance criteria:

- [x] Tested and Verified

  

`configure-gitops.sh` and `configure-pipelines.sh` should be run before `configure-dh.sh`.


scripts/configure-dh.sh
```diff
@@ -0,0 +1,363 @@
...
+EXISTING_NAMESPACE=${EXISTING_NAMESPACE:-''}
+EXISTING_DEPLOYMENT=${EXISTING_DEPLOYMENT:-''}
+EXISTING_EXTRA_ENV_SECRET=${EXISTING_EXTRA_ENV_SECRET:-''}
+RHDH_PLUGINS=${RHDH_PLUGINS:-''}
+RHDH_INSTANCE_PROVIDED=${RHDH_INSTANCE_PROVIDED:-false}
+RHDH_GITHUB_INTEGRATION=${RHDH_GITHUB_INTEGRATION:-false}
+RHDH_GITLAB_INTEGRATION=${RHDH_GITLAB_INTEGRATION:-false}
+
+# Secret variables
+GITHUB__APP__ID=${GITHUB__APP__ID:-''}
```
Jdubrick, line 26: In the case the user goes through the pipelines setup and uses the cli and enters all of the info in the prompts it should maybe set the env vars for them so they dont need to enter the info again in this script?

scripts/configure-dh.sh
```diff
@@ -0,0 +1,363 @@
...
+        if [ -z "${GITLAB__TOKEN}" ]; then
+            echo "No GitLab Token entered, try again."
+        fi
+    done
+fi
+
+# Reads Quay API Token
+# Optional: If an API Token is not entered, there will be none provided to the developer hub app config
+if [ -z "${QUAY__API_TOKEN}" ]; then
+    read -p "Enter your Quay API Token (Optional): " QUAY__API_TOKEN
```
Jdubrick, line 129: In the pipelines setup the prompt asks for Ctr+D to end this portion off but not here, is that on purpose?

catalogs.yaml
```diff
@@ -0,0 +1,3 @@
+# List of Software Template catalogs to setup with RHDH
+catalogs:
+  - https://github.com/redhat-ai-dev/ai-lab-template/blob/main/all.yaml
```
Jdubrick, line 3: Keep in mind at this current point in time until my issue is complete for swapping to this installer in the templates this will not allow component creation.

README.md
```diff
@@ -49,6 +49,12 @@ Detailed documentation for configuring GitOps/ArgoCD can be found in [`GITOPS-CO
 
 Detailed documentation for configuring Pipelines/Tekton can be found in [`PIPELINES-CONFIG.md`](./docs/PIPELINES-CONFIG.md)
 
+## Developer Hub Configuration
+
+**Note**: It is required to go through the steps under documentation mentioned in [GitOps/ArgoCD Configuration](#gitopsargocd-configuration) and [Pipelines/Tekton Configuration](#pipelinestekton-configuration) before the steps under the documentation below.
```
Jdubrick, line 64: Should we have a main script that runs all 3 in sequence?

catalogs.yaml
Jdubrick: For my testing I used because it includes the new pipeline defs alongside referencing `ai-rhdh` for argo and recognizes the proper Tekton Task for setting up the namespace: https://github.com/Jdubrick/ai-lab-template/blob/test-pipeline-definitions/all.yaml

Jdubrick (commented):
Following the installer instructions I was able to get everything setup and tested with my pipelines, you can see the results below. I still need to run through to review the manual steps as well as the pre existing.

Screenshot 2024-10-08 at 1 19 39 PM
Screenshot 2024-10-08 at 1 19 47 PM
Screenshot 2024-10-08 at 1 19 53 PM

scripts/configure-dh.sh
michael-valdron: @Jdubrick You are thinking of `QUAY__DOCKERCONFIGJSON` that is multi-line, `QUAY__API_TOKEN` is single line.

catalogs.yaml
michael-valdron: @Jdubrick Yeah you'll need to use your changes which can be done following [this](https://github.com/michael-valdron/ai-rhdh-installer/tree/feat/rhdh-config?tab=readme-ov-file#setting-catalogs-for-developer-hub-configuration), this is just the default list for when all pipeline definition and template changes are in.

scripts/configure-dh.sh
michael-valdron: > In the case the user goes through the pipelines setup and uses the cli and enters all of the info in the prompts it should maybe set the env vars for them so they dont need to enter the info again in this script?

@Jdubrick If I'm understanding correctly this is suggesting that when a user enters this information in that we should persist the information in the env vars so that they don't need to re-enter on future runs.

This is a good idea, however for the scope of this issue and the timeline I'll want to defer it to a different task.

docs/RHDH-CONFIG.md
```diff
@@ -0,0 +1,472 @@
...
+- `$EXISTING_NAMESPACE`
+  - Name of target RHDH namespace
+- `$EXISTING_DEPLOYMENT`
+  - Name of target RHDH deployment
+- `$RHDH_PLUGINS`
+  - Name of the dynamic plugins ConfigMap
+- `$EXISTING_EXTRA_ENV_SECRET`
+  - Name of the extra environment variables Secret
+
+If using a [RHDH Operator](https://github.com/redhat-developer/rhdh-operator) instance, you will need to ensure the following fields are set to the developer hub entrypoint url under any of the app configs tied to your developer hub instance:
```
Jdubrick, line 311: Is this saying you have to have a preconfigured rhdh with an app config? You can't just have a pre existing default setup?

scripts/configure-dh.sh
Jdubrick: Yeah sounds good to me!

docs/RHDH-CONFIG.md
michael-valdron: > Is this saying you have to have a preconfigured rhdh with an app config? You can't just have a pre existing default setup?

@Jdubrick This is only the case with the RHDH operator to my knowledge, the RHDH helm chart has this app config by default.

README.md
michael-valdron: Created the script in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/4aec9db72f787370738cb19095c0e089e447906a

maysunfaisal: we should probably update the docs to say to run configure.sh then

scripts/configure-dh.sh
```diff
@@ -0,0 +1,363 @@
...
+RHDH_EXTRA_ENV_SECRET="${EXTRA_ENV_SECRET}"
+RHDH_PLUGINS_CONFIGMAP="${PLUGIN_CONFIGMAP}"
+NAMESPACE=${NAMESPACE:-"ai-rhdh"}
+EXISTING_NAMESPACE=${EXISTING_NAMESPACE:-''}
+EXISTING_DEPLOYMENT=${EXISTING_DEPLOYMENT:-''}
+EXISTING_EXTRA_ENV_SECRET=${EXISTING_EXTRA_ENV_SECRET:-''}
+RHDH_PLUGINS=${RHDH_PLUGINS:-''}
+RHDH_INSTANCE_PROVIDED=${RHDH_INSTANCE_PROVIDED:-false}
+RHDH_GITHUB_INTEGRATION=${RHDH_GITHUB_INTEGRATION:-false}
+RHDH_GITLAB_INTEGRATION=${RHDH_GITLAB_INTEGRATION:-false}
```
maysunfaisal: I think both cant be defaulted to false, one should be true. And I think Github should default to true

maysunfaisal:
@michael-valdron is it possible to output these info after install, like how rhtap installer did with PLR?

1. RHDH Route URL
2. Callback URL for Github/Gitlab
3. Webhook URL for Github/Gitlab

We currently provide no messages after running helm or configure.sh script..

scripts/configure-dh.sh
maysunfaisal: We also didnt set either of these in configure.sh; so all those steps were missed when i ran configure.sh and i could not login to RHDH

scripts/configure-dh.sh
```diff
@@ -0,0 +1,363 @@
...
+
+# Reading secrets
+# Reads secrets either from environment variables or user input
+echo -n "* Reading secrets: "
+
+# Reads GitHub secrets if enabling GitHub integration
+if [[ $RHDH_GITHUB_INTEGRATION == "true" ]]; then
+    # Reads GitHub Org App ID
+    until [ ! -z "${GITHUB__APP__ID}" ]; do
+        read -p "Enter your GitHub App ID: " GITHUB__APP__ID
```
maysunfaisal, line 59: Some of these info are being asked a second time, the first time being in configure_pipelines.sh

Can we streamline this? I feel a main script like configure.sh should take all this info rather than asking twice in individual scripts..

or emphasize more on setting default-private.env before running any helm/configure scripts. Currently its after all the README about configure scripts and i missed setting this file..

scripts/configure-dh.sh
```diff
@@ -0,0 +1,363 @@
...
+    until [ ! -z "${GITHUB__APP__CLIENT__SECRET}" ]; do
+        read -p "Enter your GitHub App Client Secret: " GITHUB__APP__CLIENT__SECRET
+        if [ -z "${GITHUB__APP__CLIENT__SECRET}" ]; then
+            echo "No GitHub App Client Secret entered, try again."
+        fi
+    done
+
+    # Reads GitHub Org App Webhook URL
+    until [ ! -z "${GITHUB__APP__WEBHOOK__URL}" ]; do
+        read -p "Enter your GitHub App Webhook URL: " GITHUB__APP__WEBHOOK__URL
```
maysunfaisal: I think for better UX, we need to share how to get this, a new user would probably not know

README.md
```diff
@@ -49,6 +49,12 @@ Detailed documentation for configuring GitOps/ArgoCD can be found in [`GITOPS-CO
 
 Detailed documentation for configuring Pipelines/Tekton can be found in [`PIPELINES-CONFIG.md`](./docs/PIPELINES-CONFIG.md)
 
+## Developer Hub Configuration
```
maysunfaisal: I think we need to either

1. emphasize more on running these configuration scripts or,
2. streamline better through a main script like configure.sh

Otherwise, users will encounter errors when trying to login to RHDH via GitHub login.

I am trying to compare the UX between the rhtap-installer and this installer. rhtap-installer had two steps:

1. helm install/bash script calling helm install
2. running a PLR that was output from previous step 1 and it gives you callback url, webhook url, webhook secret, etc.

maysunfaisal:
My pipelines dont run on a custom namespace

@michael-valdron @Jdubrick should i be doing anything different or something that i missed?
Screenshot 2024-10-09 at 1 59 58 PM

Jdubrick:
> My pipelines dont run on a custom namespace
> 
> @michael-valdron @Jdubrick should i be doing anything different or something that i missed? Screenshot 2024-10-09 at 1 59 58 PM

What ai-lab template are you using? @maysunfaisal

maysunfaisal:
@Jdubrick the ones with the installer ie from templates repo main branch

Jdubrick:
> @Jdubrick the ones with the installer ie from templates repo main branch

@maysunfaisal 
Try https://github.com/Jdubrick/ai-lab-template/blob/test-pipeline-definitions/all.yaml. Because we are in a limbo state of our `main` branch in the templates supporting rhtap and this installer, the template has the pipelines + argo hardcoded into the wrong namespace.

This is because in our `main` branch of the template it tells argo to look at `rhtap`, see [here](https://github.com/redhat-ai-dev/ai-lab-template/blob/main/properties#L8-#L9), the link I sent above changes that to the deploy namespace of the installer, `ai-rhdh`, so everything will trigger properly. The branch I linked is missing some of the latest updates to the template but the pipelines and overall content is the same for testing.

This issue will be resolved with https://issues.redhat.com/browse/DEVAI-168

maysunfaisal:
@Jdubrick Thanks it works

Screenshot 2024-10-09 at 3 13 40 PM
Screenshot 2024-10-09 at 3 50 08 PM

scripts/configure-dh.sh
michael-valdron: @maysunfaisal Should be resolved with https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/40446e43b9b7d40d8f894e8eb8ee7ba8231e1b50

scripts/configure-dh.sh
michael-valdron: @maysunfaisal resolved with https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/83d0eba4b9e076fe25ea1ef8f2e127229db9353d

User is now prompted once under the `configure.sh` script instead of prompted for each script the input is used.

configure.sh
```diff
@@ -2,6 +2,120 @@
...
+    # Reads GitHub Org App Client Secret
+    until [ ! -z "${GITHUB__APP__CLIENT__SECRET}" ]; do
+        read -p "Enter your GitHub App Client Secret: " GITHUB__APP__CLIENT__SECRET
+        if [ -z "${GITHUB__APP__CLIENT__SECRET}" ]; then
+            echo "No GitHub App Client Secret entered, try again."
+        fi
+    done
+
+    # Reads GitHub Org App Webhook URL
+    until [ ! -z "${GITHUB__APP__WEBHOOK__URL}" ]; do
```
Jdubrick: If I don't run `bash ./scripts/configure-pipelines.sh` before then I don't have a webhook URL available to me to input here

maysunfaisal: Yeah I noticed that too. I just input it manually because I knew what it would be!

maysunfaisal: Why does the user need to input the webhook url env?

Jdubrick: We could probably structure the script so that the webhook url is grabbed from the secret through the script (since the pipelines config script creates this secret i believe) and then users don't need to go find it? wdyt @michael-valdron

michael-valdron: >> Why does the user need to input the webhook url env?
>
> We could probably structure the script so that the webhook url is grabbed from the secret through the script (since the pipelines config script creates this secret i believe) and then users don't need to go find it? wdyt @michael-valdron

The script could fetch the Webhook URL from the route like I did with the Backstage `baseUrl` patching.

```sh
WEBHOOK_URL="https://$(kubectl get route -n openshift-pipelines pipelines-as-code-controller --ignore-not-found -o jsonpath={.spec.host})"
```

@maysunfaisal @Jdubrick WDYT?

michael-valdron: It could even print to the console to show the user what to put into the GitHub App field.

michael-valdron: > since the pipelines config script creates this secret i believe

No, the pipelines config script creates a secret that stores the `GITHUB__APP__WEBHOOK__SECRET` not the `GITHUB__APP__WEBHOOK__URL`.

Jdubrick: +1 I like the idea of fetching it within the `configure.sh` if you decide to use that instead of running the 3 setup scripts manually, and printing it out will also be helpful!

Does it not create the secret containing the url here? https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/main/scripts/configure-pipelines.sh#L139-#L143

@michael-valdron

maysunfaisal: I think we need to print this info https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10#issuecomment-2402874415 because users will need to also update their GitHub application. A total new person will not know how to do this unless installer guides them

michael-valdron: > Does it not create the secret containing the url here? https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/main/scripts/configure-pipelines.sh#L139-#L143
> 
> @michael-valdron

@Jdubrick Ah you right it is already being fetched on [line 143](https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/753af9a15690059aa3f1f2eba5ce6d5f7efa671e/scripts/configure-pipelines.sh#L143) and set into the secret, didn't notice that my bad.

> +1 I like the idea of fetching it within the `configure.sh` if you decide to use that instead of running the 3 setup scripts manually, and printing it out will also be helpful!

With the above, I'm thinking the quickest way to do this is I can add this step in both pipelines and rhdh config scripts with control logic to determine if the secret has been created yet, if so just fetch the webhook url from the secret and if not create the secret as it currently does.

michael-valdron: > I think we need to print this info [#10 (comment)](https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10#issuecomment-2402874415) because users will need to also update their GitHub application. A total new person will not know how to do this unless installer guides them

Yeah I agree that anything we need from the cluster should be printed out so the user can use for the GitHub app fields.

Jdubrick: +1 to both points! I like those ideas

README.md
michael-valdron: Created doc section in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/c9c41273c392f968aebe98f7919a368ed6e2923d

scripts/configure-dh.sh
michael-valdron: @maysunfaisal User prompting for this is no longer the case with https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/d14cd482e6b67a1f54fe53fd3c76e78a5a81c4df, scripts now fetches this from either the rhdh env var secret or from the route itself if the secret does not contain this variable. User can still override the env var if they choose.

configure.sh
```diff
@@ -0,0 +1,139 @@
...
+echo "**GitOps/ArgoCD Configuration**"
+
+bash $BASE_DIR/scripts/configure-gitops.sh
+if [ $? -ne 0 ]; then
+    echo "GitOps/ArgoCD Configuration: FAILED"
+    exit 1
+fi
+echo "GitOps/ArgoCD Configuration: OK"
+
+echo "**Pipelines/Tekton Configuration**"
```
Jdubrick, line 125: Somewhere around here I believe it is moving too fast before resources are created, trying to run `configure.sh` to configure everything I run into:

```
**GitOps/ArgoCD Configuration**
configmap/argocd-config created
GitOps/ArgoCD Configuration: OK
**Pipelines/Tekton Configuration**
* Reading secrets: 
OK
* Waiting for 'tektonconfigs' CRD: OK
* Waiting for pipelines operator deployment: OK
* Fetching Webhook URL: error: the server doesn't have a resource type "ai-rh-developer-hub-env"
Extra environment variable secret 'ai-rh-developer-hub-env' not found!FAIL
Pipelines/Tekton Configuration: FAILED
```

README.md
```diff
@@ -41,14 +41,28 @@ Run `helm upgrade --install <release-name> <path-to-chart> --namespace <namespac
 
 This installer is incompatible with `default` namespace installations, install and uninstall commands must include `--namespace <target-namespace>` or the context namespace must be changed, e.g. `oc project <target-namespace>`.
 
-## GitOps/ArgoCD Configuration
+## Configuration
+
+To configure your RHDH/GitOps/Pipelines setup to use AI Software Templates, run the following:
+
+`NAMESPACE=<target-namespace> bash ./configure.sh`
```
Jdubrick: Nit: If you don't set namespace and it auto defaults to `ai-rhdh` for example we should outline that here so that people don't need to set it if using default

configure.sh
Jdubrick: Interestingly enough though the secret is here:
Screenshot 2024-10-15 at 11 37 53 AM

michael-valdron: fixed typos in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/2f9780108f70d731f72a87309e3bd32acc00862c

README.md
michael-valdron: @Jdubrick Outlined this in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/f79731d79bd8ba133e35bab0b0ec09d8dcc70445 changes.

scripts/configure-pipelines.sh
```diff
@@ -117,12 +133,49 @@ if [ $? -ne 0 ]; then
 fi
 echo "OK"
 
+# Fetching Webhook URL
+if [[ $RHDH_GITHUB_INTEGRATION == "true" ]]; then
+    echo -n "* Fetching Webhook URL: "
```
Jdubrick, line 138: Currently running into the following:
```
**GitOps/ArgoCD Configuration**
configmap/argocd-config created
GitOps/ArgoCD Configuration: OK
**Pipelines/Tekton Configuration**
* Reading secrets: 
OK
* Waiting for 'tektonconfigs' CRD: OK
* Waiting for pipelines operator deployment: OK
* Fetching Webhook URL: Error: 1:35: invalid input text "aHR0cHM6Ly9waXBl..."
Error from server (BadRequest): error decoding patch: json: cannot unmarshal array into Go value of type map[string]interface {}
error: the server doesn't have a resource type "ai-rh-developer-hub-env"
FAIL
Pipelines/Tekton Configuration: FAILED
```

Is it possible something being fetched isn't being decoded and we are passing b64 somewhere?
@michael-valdron

michael-valdron: @Jdubrick Found a bit more typos in the `configure-pipelines.sh` script and re-wrote the logic there for patching the secret in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/dcf9450c78e6fa4bae3957fc77b4d1b09243a39a. Originally wrote it in the flow of `kubectl apply` and realized piping the patch content into `kubectl patch secret ${RHDH_EXTRA_ENV_SECRET} -n $NAMESPACE --type 'merge' -p - >/dev/null` does not work, needed to use a variable instead.

michael-valdron: Tested the current state of the scripts on my cluster and seems to work now.

Jdubrick: Can confirm `configure.sh` is working for me now, I know we talked about it earlier in a different review thread but will we be adding an output that contains the homepage/callback/webhook urls? @michael-valdron

README.md
```diff
@@ -41,14 +41,30 @@ Run `helm upgrade --install <release-name> <path-to-chart> --namespace <namespac
 
 This installer is incompatible with `default` namespace installations, install and uninstall commands must include `--namespace <target-namespace>` or the context namespace must be changed, e.g. `oc project <target-namespace>`.
 
-## GitOps/ArgoCD Configuration
+## Configuration
+
+To configure your RHDH/GitOps/Pipelines setup to use AI Software Templates, run the following:
```
Jdubrick:
```suggestion
For convenience you can configure your Red Hat Developer Hub, GitOps, and Pipelines Operators to enable the use of AI Software Templates by running the following:
```

README.md
```diff
@@ -41,14 +41,30 @@ Run `helm upgrade --install <release-name> <path-to-chart> --namespace <namespac
...
+
+`bash ./configure.sh`
+
+Please note if you changed the installation namespace used by the installer you will first need to run `export NAMESPACE=<namespace used>` as the default value is `ai-rhdh`.
```
Jdubrick:
```suggestion
**Note**: If you changed the installation namespace used by the installer you will first need to run `export NAMESPACE=<namespace used>` as the default value is `ai-rhdh`.
```

README.md
```diff
@@ -41,14 +41,30 @@ Run `helm upgrade --install <release-name> <path-to-chart> --namespace <namespac
...
+
+See the documents below on the configuration details for setting up each component.
```
Jdubrick:
```suggestion
Alternatively, if you do not wish to use the all-in-one `configure.sh` script, you can find documentation below for configuring each component individually.
```

Jdubrick (commented):
Couple small comments, is there anything else major you are going to commit to this PR that I will need to re-review? @michael-valdron

scripts/configure-pipelines.sh
michael-valdron: > I know we talked about it earlier in a different review thread but will we be adding an output that contains the homepage/callback/webhook urls? @michael-valdron

@Jdubrick Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/3a820df5e477b1d8c80c23fb8bdd2951e9527afe

configure.sh
michael-valdron: > > I think we need to print this info [#10 (comment)](https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10#issuecomment-2402874415) because users will need to also update their GitHub application. A total new person will not know how to do this unless installer guides them
> 
> Yeah I agree that anything we need from the cluster should be printed out so the user can use for the GitHub app fields.

Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/3a820df5e477b1d8c80c23fb8bdd2951e9527afe

FYI @maysunfaisal

README.md
michael-valdron: > running a PLR that was output from previous step 1 and it gives you callback url, webhook url, webhook secret, etc.

@maysunfaisal  Done in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/3a820df5e477b1d8c80c23fb8bdd2951e9527afe

michael-valdron: > streamline better through a main script like configure.sh

@maysunfaisal Provided in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/4aec9db72f787370738cb19095c0e089e447906a

michael-valdron: @maysunfaisal Is there any other UX improvements that should be included in these changes?

configure.sh
michael-valdron: > > Does it not create the secret containing the url here? https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/main/scripts/configure-pipelines.sh#L139-#L143
> > @michael-valdron
> 
> @Jdubrick Ah you right it is already being fetched on [line 143](https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/753af9a15690059aa3f1f2eba5ce6d5f7efa671e/scripts/configure-pipelines.sh#L143) and set into the secret, didn't notice that my bad.
> 
> > +1 I like the idea of fetching it within the `configure.sh` if you decide to use that instead of running the 3 setup scripts manually, and printing it out will also be helpful!
> 
> With the above, I'm thinking the quickest way to do this is I can add this step in both pipelines and rhdh config scripts with control logic to determine if the secret has been created yet, if so just fetch the webhook url from the secret and if not create the secret as it currently does.

Provided in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/d14cd482e6b67a1f54fe53fd3c76e78a5a81c4df, with a fixed control logic in https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/10/commits/4ec8f7a3b170dae6eb33f22e7c49abff4bb48e7b

FYI @maysunfaisal @Jdubrick

scripts/install-info.sh
```diff
@@ -0,0 +1,139 @@
...
+
+# Constants
+GITHUB_DOCS_URL='https://pipelinesascode.com/docs/install/github_apps/'
+DEFAULT_RHDH_DEPLOYMENT="backstage-ai-rh-developer-hub" # deployment created by rhdh operator by default
+EXTRA_ENV_SECRET="ai-rh-developer-hub-env" # secret created by rhdh installer to store private env vars
+
+# Variables
+BASE_DIR="$(realpath $(dirname ${BASH_SOURCE[0]}))"
+RHDH_DEPLOYMENT="${DEFAULT_RHDH_DEPLOYMENT}"
+RHDH_ROUTE="${RHDH_ROUTE:-$RHDH_DEPLOYMENT}"
```
Jdubrick, line 11: Nit: Do we need this intermediate var?

```suggestion
RHDH_ROUTE="${RHDH_ROUTE:-$DEFAULT_RHDH_DEPLOYMENT}"
```

Jdubrick: Also, am I missing where this is set to something other than the default for those who that use case applies to?

Jdubrick (commented):
Currently restarting my cluster to run through and see how the new print script works alongside the config, in the meantime left some general comments/questions I have

maysunfaisal (approved):
/lgtm

Thanks for the updates. Tried it out and it looks awesome.

maysunfaisal:
@michael-valdron Please squash merge when merging

README.md
maysunfaisal: it looks good now, thx

scripts/install-info.sh
michael-valdron: > Nit: Do we need this intermediate var?

@Jdubrick `RHDH_ROUTE="${RHDH_ROUTE:-$RHDH_DEPLOYMENT}"` is correct, if the RHDH deployment name is set to something else, the route name should be too.

Jdubrick (approved):
lgtm, using the `configure.sh` with the output is much easier on the user now (and that's coming from me who knows how to set it up in any way! lol)

PR #34, opened 2024-11-08
[DEVAI-231] Create Private.env After Manual Input

Jdubrick:
### What does this PR do?:

This PR adds logic to create the `private.env` file after a user does the manual input so they only have to do it once. It also sources (if exists) the `private.env` at the beginning of `configure.sh` to remove the responsibility from the user for doing that. I also added the GitHub/GitLab environment variables to the list of exports so a user can add that to their file since it was missing.

Contains doc updates to reflect these changes.

### PR acceptance criteria:

- [x] Tested and Verified

  

I tested the following scenarios:
- No `private.env` file, input info via CLI and everything deployed as expected. `private.env` file was created for me.
- Used the created `private.env` file to reconfigure after an uninstall, everything deployed as expected.
- Tried a partially filled out `private.env` file, was only prompted for the missing values. `private.env` file was updated for me and everything deployed as expected.


johnmcollier (approved):
LGTM

PR #35, opened 2024-11-20
Enable gitlab login with gilab integration

jrichter1:
### What does this PR do?:
Switches the sign in page to gitlab provider, if gitlab integration is configured. Otherwise keeps github sign in.

### PR acceptance criteria:

- [ ] Tested and Verified

  


Jdubrick (commented):
Will we run into conflicts if _both_ `RHDH_GITHUB_INTEGRATION` and `RHDH_GITLAB_INTEGRATION` are enabled? All of the other fields we update have the explicit `.github` or `.gitlab` naming but `signInPage` does not?

I think we should add a piece to the docs too for manual implementations, namely [here](https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/main/docs/RHDH-CONFIG.md#ai-rhdh-installer-manual-configuration) and [here](https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/main/docs/RHDH-CONFIG.md#wip-pre-existing-instance-manual-configuration).

jrichter1:
> Will we run into conflicts if _both_ `RHDH_GITHUB_INTEGRATION` and `RHDH_GITLAB_INTEGRATION` are enabled? All of the other fields we update have the explicit `.github` or `.gitlab` naming but `signInPage` does not?
> 

gitlab wins if both are present, but only because it gets configured last. for some reason it's one xor the other

Jdubrick:
> > Will we run into conflicts if _both_ `RHDH_GITHUB_INTEGRATION` and `RHDH_GITLAB_INTEGRATION` are enabled? All of the other fields we update have the explicit `.github` or `.gitlab` naming but `signInPage` does not?
> 
> gitlab wins if both are present, but only because it gets configured last. for some reason it's one xor the other

Okay, can we add the required lines you need to set to the manual sections I linked above and add a snippet letting users know that if both integrations are enabled GitLab is going to win out? Seems like a Backstage/RHDH limitation that we should note?

jrichter1:
I've improved it a bit, letting the user choose when both providers are set up, or just override it with an environment variable

configure.sh
```diff
@@ -106,6 +110,32 @@ if [[ $RHDH_GITLAB_INTEGRATION == "true" ]]; then
             echo "No GitLab Token entered, try again."
         fi
     done
+
+    # Choose which sign in provider to use
+    if [[ $RHDH_GITHUB_INTEGRATION == "true" ]] && [[ ! $RHDH_SIGNIN_PROVIDER =~ ^(github|gitlab)$ ]]; then
+        echo "Multiple authentication providers detected"
+        PS3='Select the desired sign in method: '
```
Jdubrick: When trying this out myself it displayed as:

1) github
2) gitlab

It wasn't super clear if I should be entering the number or the word itself, or either? My only comment would be if we can make this prompt more clear

Jdubrick (commented):
Just one small comment I thought of when running through the workflow myself. Also since we will be supporting GitLab login with this can we also remove this notice?

https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/main/README.md?plain=1#L4

Jdubrick:
> Just one small comment I thought of when running through the workflow myself. Also since we will be supporting GitLab login with this can we also remove this notice?
> 
> https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/main/README.md?plain=1#L4

@jrichter1 https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/9 was also merged so we will need a rebase as well

configure.sh
jrichter1: it now allows both, and should tell you so

Jdubrick (approved):
lgtm, @michael-valdron any additional comments?

PR #39, opened 2025-02-05
RHDHPAI-36: Swap ArgoCD Frontend Plugin

Jdubrick:
### What does this PR do?:

This PR swaps the ArgoCD frontend plugin from `./dynamic-plugins/dist/roadiehq-backstage-plugin-argo-cd` to `./dynamic-plugins/dist/backstage-community-plugin-redhat-argocd` as the `backstage-community-plugin-redhat-argocd` plugin is preinstalled as part of rhdh v1.3 and above. 

Additionally, this PR contains a fix for a `context exceeded` error that is randomly occurring when configuring ArgoCD. You can see in the link above the new `skip-tls` field that even the official ArgoCD docs reference this as a fix. By adding it to the `values.yaml` it gives users the ability to toggle this functionality if required.

https://issues.redhat.com/browse/RHDHPAI-36

### PR acceptance criteria:

- [x] Tested and Verified

  

You will need to perform the normal rhdh install to a cluster, keep in mind since this is rhdh 1.4 and https://issues.redhat.com/browse/RHDHPAI-359 is not complete yet you will need to make sure _all_ GitHub plugins are installed for the templates to work (if you want to see the app deploy and the plugin to work).

You can then view the changes for ArgoCD.

I have included screenshots below of the plugin as well.

Screenshot 2025-02-05 at 2 51 33 PM
Screenshot 2025-02-05 at 3 00 09 PM
Screenshot 2025-02-05 at 3 00 15 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,10 +28,6 @@ This Helm chart installs and configures the following operators:
 
 **Note**: If a subscription for an operator already exists, the installation will not tamper with it.
 
-### Helm Troubleshooting
-
-If you find the `ArgoCD Operator` is installed but the configuration is failing due to `context exceeded`, please ensure that `skip-test-tls` is set to `true` in your [values.yaml](./chart/values.yaml) file.
-
 ### Install
 
 >[!WARNING]
@@ -52,6 +48,15 @@ helm upgrade --install ai-rhdh ./chart --namespace ai-rhdh --create-namespace
 To uninstall the Operators, run:
 ```
 helm uninstall <release-name> --namespace <namespace>
+```
+
+### Helm Troubleshooting
+
+If you find the `ArgoCD Operator` is installed but the configuration is failing due to `context exceeded`, please ensure that `skip-test-tls` is set to `true` in your [values.yaml](./chart/values.yaml) file.
+
+You can pass the `skip-test-tls` value as a flag in your Helm upgrade command as well if you prefer:
+```
+helm upgrade --install ai-rhdh ./chart --namespace ai-rhdh --create-namespace --set openshift-gitops.skip-test-tls=true
 ```
 
 ## Configuration
--- a/README.md
+++ b/README.md
@@ -28,6 +28,10 @@ This Helm chart installs and configures the following operators:
 
 **Note**: If a subscription for an operator already exists, the installation will not tamper with it.
 
+### Helm Troubleshooting
+
+If you find the `ArgoCD Operator` is installed but the configuration is failing due to `context exceeded`, please ensure that `skip-tls` is set to `true` in your [values.yaml](./chart/values.yaml) file.
+
 ### Install
 
 >[!WARNING]
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ This Helm chart installs and configures the following operators:
 
 ### Helm Troubleshooting
 
-If you find the `ArgoCD Operator` is installed but the configuration is failing due to `context exceeded`, please ensure that `skip-tls` is set to `true` in your [values.yaml](./chart/values.yaml) file.
+If you find the `ArgoCD Operator` is installed but the configuration is failing due to `context exceeded`, please ensure that `skip-test-tls` is set to `true` in your [values.yaml](./chart/values.yaml) file.
 
 ### Install
 
--- a/chart/templates/openshift-gitops/includes/_configure.tpl
+++ b/chart/templates/openshift-gitops/includes/_configure.tpl
@@ -59,7 +59,7 @@
           ARGOCD_PASSWORD="$(kubectl get secret -n "$NAMESPACE" "ai-$RHDH_ARGOCD_INSTANCE-cluster" -o jsonpath="{.data.admin\.password}" | base64 --decode)"
           echo -n "."
           RETRY=0
-          while ! ./argocd login "$ARGOCD_HOSTNAME" --grpc-web --insecure --http-retry-max 5 --username admin --password "$ARGOCD_PASSWORD"{{- if index .Values "openshift-gitops" "skip-tls" }} --skip-test-tls{{- end }}>/dev/null; do
+          while ! ./argocd login "$ARGOCD_HOSTNAME" --grpc-web --insecure --http-retry-max 5 --username admin --password "$ARGOCD_PASSWORD"{{- if eq (index .Values "openshift-gitops" "skip-test-tls") "true" }} --skip-test-tls{{- end }}>/dev/null; do
             if [ "$RETRY" = "20" ]; then
               echo "FAIL"
               echo "[ERROR] Could not login to ArgoCD" >&2
--- a/chart/templates/openshift-gitops/includes/_configure.tpl
+++ b/chart/templates/openshift-gitops/includes/_configure.tpl
@@ -59,7 +59,7 @@
           ARGOCD_PASSWORD="$(kubectl get secret -n "$NAMESPACE" "ai-$RHDH_ARGOCD_INSTANCE-cluster" -o jsonpath="{.data.admin\.password}" | base64 --decode)"
           echo -n "."
           RETRY=0
-          while ! ./argocd login "$ARGOCD_HOSTNAME" --grpc-web --insecure --http-retry-max 5 --username admin --password "$ARGOCD_PASSWORD" >/dev/null; do
+          while ! ./argocd login "$ARGOCD_HOSTNAME" --grpc-web --insecure --http-retry-max 5 --username admin --password "$ARGOCD_PASSWORD"{{- if index .Values "openshift-gitops" "skip-tls" }} --skip-test-tls{{- end }}>/dev/null; do
             if [ "$RETRY" = "20" ]; then
               echo "FAIL"
               echo "[ERROR] Could not login to ArgoCD" >&2
--- a/chart/templates/openshift-gitops/includes/_configure.tpl
+++ b/chart/templates/openshift-gitops/includes/_configure.tpl
@@ -59,7 +59,7 @@
           ARGOCD_PASSWORD="$(kubectl get secret -n "$NAMESPACE" "ai-$RHDH_ARGOCD_INSTANCE-cluster" -o jsonpath="{.data.admin\.password}" | base64 --decode)"
           echo -n "."
           RETRY=0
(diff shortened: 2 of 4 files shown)


chart/values.yaml
```diff
@@ -17,6 +17,8 @@ openshift-gitops:
         requests:
           cpu: "1"
           memory: 3Gi
+  # TLS check can cause unexpected hanging during login: https://argo-cd.readthedocs.io/en/stable/user-guide/commands/argocd_login/#options
+  skip-tls: true
```
michael-valdron: Despite the current state of ArgoCD for us, I'd still make the default `false`:
```suggestion
  skip-tls: false
```

chart/templates/openshift-gitops/includes/_configure.tpl
```diff
@@ -59,7 +59,7 @@
           ARGOCD_PASSWORD="$(kubectl get secret -n "$NAMESPACE" "ai-$RHDH_ARGOCD_INSTANCE-cluster" -o jsonpath="{.data.admin\.password}" | base64 --decode)"
           echo -n "."
           RETRY=0
-          while ! ./argocd login "$ARGOCD_HOSTNAME" --grpc-web --insecure --http-retry-max 5 --username admin --password "$ARGOCD_PASSWORD" >/dev/null; do
+          while ! ./argocd login "$ARGOCD_HOSTNAME" --grpc-web --insecure --http-retry-max 5 --username admin --password "$ARGOCD_PASSWORD"{{- if index .Values "openshift-gitops" "skip-tls" }} --skip-test-tls{{- end }}>/dev/null; do
```
michael-valdron: I would use `eq` here, I found through additional testing that it can still be `true` if its parsed as the string `"false"` (even if in yaml its a boolean):

```suggestion
          while ! ./argocd login "$ARGOCD_HOSTNAME" --grpc-web --insecure --http-retry-max 5 --username admin --password "$ARGOCD_PASSWORD"{{- if eq (index .Values "openshift-gitops" "skip-tls") "true" }} --skip-test-tls{{- end }}>/dev/null; do
```

We should fix this in the other places as well.

michael-valdron (changes requested):
lgtm after a few minor feedback is addressed.

Tested and changes with my [WIP changes](https://github.com/michael-valdron/ai-rhdh-installer/tree/chore/migrate-dynamic-plugins) that includes the new replace community plugins for the broken `idp-janis` plugins works as expected:

Screenshot from 2025-02-05 19-09-37
Screenshot from 2025-02-05 19-27-37
Screenshot from 2025-02-05 19-28-08
Screenshot from 2025-02-05 19-28-52

thepetk (commented):
In general looks good to me after @michael-valdron comments are addressed. Just a minor comment, I feel that it'd be better to keep `skip-test-tls` as the name for the `values.yaml` field because the current name gives the impression that we are disabling the TLS in general, while we only skip the tests on the argoCD side, if TLS is enabled.

thepetk:
Looks good to me in general. The fix with `skip-test-tls` worked for me. One note is it worked both on `1.4` and `1.2` so a question is if we are going to backport this fix. Apart from that what do you think about using the `skip-test-tls` as a fallback in case the `argocd admin` phase fails?

We could apply it if the normal flow fails n number of times. WDYT?

chart/templates/openshift-gitops/includes/_configure.tpl
Jdubrick: Good catch, what other places are you referring to? Just in general for the installer where we do checks?

Jdubrick:
> Looks good to me in general. The fix with `skip-test-tls` worked for me. One note is it worked both on `1.4` and `1.2` so a question is if we are going to backport this fix. Apart from that what do you think about using the `skip-test-tls` as a fallback in case the `argocd admin` phase fails?
> 
> We could apply it if the normal flow fails n number of times. WDYT?

@michael-valdron @thepetk I've addressed the comments you both left in my latest commit, primarily the naming, default value, and equality check. As for having it as the fallback incase this fails I am open to hearing what you both think - In my mind it would be a nice quality of life, but I think having it by default be the fallback might take the control out of the users' hands when it comes to if they want to skip that test or not.

In other words, we have the toggle in place in the `values.yaml`, so if we are planning on using it as the fallback we may as well not have that control if that makes sense?

chart/templates/openshift-gitops/includes/_configure.tpl
Jdubrick: Need to investigate this further - Running into an issue where the comparison type is not the same currently

thepetk:
> > Looks good to me in general. The fix with `skip-test-tls` worked for me. One note is it worked both on `1.4` and `1.2` so a question is if we are going to backport this fix. Apart from that what do you think about using the `skip-test-tls` as a fallback in case the `argocd admin` phase fails?
> > We could apply it if the normal flow fails n number of times. WDYT?
> 
> @michael-valdron @thepetk I've addressed the comments you both left in my latest commit, primarily the naming, default value, and equality check. As for having it as the fallback incase this fails I am open to hearing what you both think - In my mind it would be a nice quality of life, but I think having it by default be the fallback might take the control out of the users' hands when it comes to if they want to skip that test or not.
> 
> In other words, we have the toggle in place in the `values.yaml`, so if we are planning on using it as the fallback we may as well not have that control if that makes sense?

Yeah +1 at least this is what I have in mind too. When `skip-test-tls` is `true` we allow n number of attempts if failed and then we fallback to skip the tests. If false we don't use the fallback. This way I feel we at least try the "correct" way (with tests included) at least a couple of times before we skip them.

thepetk (approved):
lgtm. worked for my tests and I had to use it to install 1.2 too.

chart/templates/openshift-gitops/includes/_configure.tpl
```diff
@@ -59,7 +59,7 @@
           ARGOCD_PASSWORD="$(kubectl get secret -n "$NAMESPACE" "ai-$RHDH_ARGOCD_INSTANCE-cluster" -o jsonpath="{.data.admin\.password}" | base64 --decode)"
           echo -n "."
           RETRY=0
-          while ! ./argocd login "$ARGOCD_HOSTNAME" --grpc-web --insecure --http-retry-max 5 --username admin --password "$ARGOCD_PASSWORD" >/dev/null; do
+          while ! ./argocd login "$ARGOCD_HOSTNAME" --grpc-web --insecure --http-retry-max 5 --username admin --password "$ARGOCD_PASSWORD"{{- if eq (index .Values "openshift-gitops" "skip-test-tls") "true" }} --skip-test-tls{{- end }}>/dev/null; do
```
michael-valdron: @Jdubrick Running into the comparison error you [mentioned](https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/39#discussion_r1947110959) before:

```
Error: template: ai-rhdh-setup/templates/configure.yaml:27:10: executing "ai-rhdh-setup/templates/configure.yaml" at <include "rhdh.gitops.configure" .>: error calling include: template: ai-rhdh-setup/templates/openshift-gitops/includes/_configure.tpl:62:146: executing "rhdh.gitops.configure" at <eq (index .Values "openshift-gitops" "skip-test-tls") "true">: error calling eq: incompatible types for comparison
```

Removing the quotes on `true` fixes this error:

```suggestion
          while ! ./argocd login "$ARGOCD_HOSTNAME" --grpc-web --insecure --http-retry-max 5 --username admin --password "$ARGOCD_PASSWORD"{{- if eq (index .Values "openshift-gitops" "skip-test-tls") true }} --skip-test-tls{{- end }}>/dev/null; do
```

Jdubrick:
@thepetk @michael-valdron I swapped the check to look for `true` as a boolean instead of the string `"true"`. I think in this case it being an explicit boolean is better suited. 

As for the discussion around auto-disabling the test-tls check if it fails N times I think it would be good for a parking lot/opening a spike for it, wdyt?

README.md
```diff
@@ -28,6 +28,10 @@ This Helm chart installs and configures the following operators:
 
 **Note**: If a subscription for an operator already exists, the installation will not tamper with it.
 
+### Helm Troubleshooting
+
+If you find the `ArgoCD Operator` is installed but the configuration is failing due to `context exceeded`, please ensure that `skip-test-tls` is set to `true` in your [values.yaml](./chart/values.yaml) file.
+
```
michael-valdron: It would be good to provide an example of this added as a flag:

`helm upgrade --install <release-name> <path-to-chart> --namespace <namespace> --create-namespace --set openshift-gitops.skip-test-tls=true`

README.md
```diff
@@ -28,6 +28,10 @@ This Helm chart installs and configures the following operators:
 
 **Note**: If a subscription for an operator already exists, the installation will not tamper with it.
 
+### Helm Troubleshooting
```
michael-valdron: With my other comment might make sense to change the ordering to be between **uninstall** and **configuration** sections.

michael-valdron (changes requested):

chart/templates/openshift-gitops/includes/_configure.tpl
michael-valdron: Beat me to my suggestion :wink:

michael-valdron:
> As for the discussion around auto-disabling the test-tls check if it fails N times I think it would be good for a parking lot/opening a spike for it, wdyt?

@Jdubrick  Yeah opening a spike to investigate auto disabling when fail N times sounds good, :+1:, I would expand in suggestion it should be explicitly fail timeout N times rather than any failure if possible.

thepetk:
> > As for the discussion around auto-disabling the test-tls check if it fails N times I think it would be good for a parking lot/opening a spike for it, wdyt?
> 
> @Jdubrick Yeah opening a spike to investigate auto disabling when fail N times sounds good, 👍, I would expand in suggestion it should be explicitly fail timeout N times rather than any failure if possible.

I like the timeout approach too!

Jdubrick:
PR comments addressed and opened a spike issue for the timeout handling: https://issues.redhat.com/browse/RHDHPAI-567

michael-valdron (approved):
/lgtm

thepetk (approved):

PR #40, opened 2025-02-11
[RHDHPAI-312] Add Kubernetes Service Account Token

Jdubrick:
### What does this PR do?:

This PR adds a new definition to the Kubernetes resources being created by the Helm chart to allow for a Kubernetes Service Account token to be created and linked to the SA. This is required because in later versions of Kubernetes SA's don't automatically generate their tokens. This change will not affect the functionality of the installer when deploying to OCP versions that _do_ automatically generate that token as it will just reference the manually created one.

This was tested with RHDH `v1.4` on all of its supported OCP versions, `v4.14`, `v4.15`, `v4.16`, and `v4.17`.

https://issues.redhat.com/browse/RHDHPAI-312

### PR acceptance criteria:

- [x] Tested and Verified

  

Run the installer on any of the supported OCP versions and navigate to the `Kubernetes` and `Topology` tabs of a deployed Component, you will no longer see the error being presented and instead can see it working as expected.

### DIFF
--- a/docs/pipelines/PRE-EXISTING.md
+++ b/docs/pipelines/PRE-EXISTING.md
@@ -76,7 +76,7 @@ First you will need to create the [service account](../../chart/templates/k8s-se
 that RHDH will need to interact with the cluster such as
 creating tekton pipeline runs.
 
-Once the service account is created there will be a tied secret which stores the service account token, e.g. if a service account `rhdh-kubernetes-plugin` is created then a secret with the name pattern `rhdh-kubernetes-plugin-token-*` is also created.
+Once the service account is created there will be a tied secret which stores the service account token, e.g. if a service account `rhdh-kubernetes-plugin` is created then a secret with the name `rhdh-kubernetes-plugin-token` is also created.
 
 Keep note of the name of this secret.
 
--- a/docs/pipelines/PRE-EXISTING.md
+++ b/docs/pipelines/PRE-EXISTING.md
@@ -72,13 +72,9 @@ Once you have done the prior steps and have the information from the prior steps
 You will follow the same steps as [step 3 of the script configuration for a pre-existing instance](#step-3-configure-cosign).
 
 #### Step 2: Kubernetes API Service Account
-First you will need to create the [service account](../../chart/templates/k8s-serviceaccount.yaml)
-that RHDH will need to interact with the cluster such as
-creating tekton pipeline runs.
+You will first need to create a Kubernetes Service Account. You can reference [`KUBERNETES_SERVICEACCOUNT.md`](./KUBERNETES_SERVICEACCOUNT.md) for more information on how to create these resources.
 
-Once the service account is created there will be a tied secret which stores the service account token, e.g. if a service account `rhdh-kubernetes-plugin` is created then a secret with the name `rhdh-kubernetes-plugin-token` is also created.
-
-Keep note of the name of this secret.
+It is important to keep in mind the name of both the Service Account and the Token Secret.
 
 #### Step 3: Create App Namespace Setup Task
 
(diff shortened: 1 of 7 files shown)


docs/pipelines/PRE-EXISTING.md
```diff
@@ -76,7 +76,7 @@ First you will need to create the [service account](../../chart/templates/k8s-se
 that RHDH will need to interact with the cluster such as
```
michael-valdron: Based on past feedback I think I'll be good to create a document that explicitly shows how to create the Kubernetes service account and tied resources, then reference it here and under the RHDH pre-existing guide.

docs/pipelines/PRE-EXISTING.md
```diff
@@ -76,7 +76,7 @@ First you will need to create the [service account](../../chart/templates/k8s-se
...
 creating tekton pipeline runs.
 
-Once the service account is created there will be a tied secret which stores the service account token, e.g. if a service account `rhdh-kubernetes-plugin` is created then a secret with the name pattern `rhdh-kubernetes-plugin-token-*` is also created.
+Once the service account is created there will be a tied secret which stores the service account token, e.g. if a service account `rhdh-kubernetes-plugin` is created then a secret with the name `rhdh-kubernetes-plugin-token` is also created.
```
michael-valdron: Since the installer is creating the service account and tied secret it should not be assumed that the secret is created with the service account, the previous comment highlights what should be done to instruct the user to create these resources.

Here we should just state to keep note of the name of the secret created for later steps.

michael-valdron (commented):
Other than feedback looks good to me, tested with installer and Kubernetes plugin is initializing now without error.

Final thought I had with instruction and configure script is that we assume that the service account and token secret will be created with the same name as the installer provided (that is constant). Though this could a follow up issue, should we not assume this and add an override (at least) in case its named something else?

Jdubrick:
@michael-valdron I added snippets for instructing users on creating the resources, I think we can open a separate issue for allowing the override, I can open that today and will link it with this issue. It does make me question however with the override just how many environment variables we are supporting now for the various use cases, I wonder if there is any way to reduce the amount of information required?

Jdubrick:
Follow up issue for the Kubernetes SA/token override opened: https://issues.redhat.com/browse/RHDHPAI-587

docs/rhdh/PRE-EXISTING.md
```diff
@@ -67,14 +67,20 @@ Once you have done the prior steps and have the information from the prior steps
...
 ### Manual Configuration
 
-#### Step 1: Create Extra Environment Variables Secret
+#### Step 1: Create Kubernetes Service Account and Token
+
+You can skip this step if these resources are already present in your desired namespace.
+
+For information related to creating these resources see [`KUBERNETES_SERVICEACCOUNT.md`](../pipelines/KUBERNETES_SERVICEACCOUNT.md).
+
+#### Step 2: Create Extra Environment Variables Secret
```
michael-valdron, line 76: Creating this new step 2 should have the following steps updated as well, and references to these sections.

michael-valdron (commented):
@Jdubrick Your follow up changes looks good to me, just the matter of the change in section formatting to be addressed.

> It does make me question however with the override just how many environment variables we are supporting now for the various use cases, I wonder if there is any way to reduce the amount of information required?

I think what was discussed in our team meeting this morning the planned future of installing RHDH instances for AI Software Templates should take care of this. But for the context of the overrides these would not be required fields rather optional fields to give this choice to the individual using the scripts. FWIW the nature of this would not see this to be a high priority atm but still good to have an issue open for it.

michael-valdron (approved):
/lgtm

PR #41, opened 2025-02-13
RHDHPAI-104: Wait for RHDH deployments to finish updating

michael-valdron:
### What does this PR do?:

Waits for RHDH deployments to finish updating during the installation. This will wait to patch the deployment with the route generated hostname until the deployment is finished initializing, along with waiting to complete the installation RHDH configure job until that same patch is finished too. This should prevent resource deadlocks from happening.

https://issues.redhat.com/browse/RHDHPAI-104

### PR acceptance criteria:

- [X] Tested and Verified

  

1. Set the `--skip-test-tls` flag [here](https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/0542e44be17653cdebfe6a18ed253acc6edec62b/chart/templates/openshift-gitops/includes/_configure.tpl#L62) to prevent issue that was fixed in the v1.4 migration changes: https://github.com/redhat-ai-dev/ai-rhdh-installer/commit/5ba9c63e05b2c5780e32165f5857af1c0540a076
2. Try installing with the helm chart: `helm upgrade --install ai-rhdh ./chart --namespace ai-rhdh --create-namespace`

### DIFF
--- a/chart/templates/developer-hub/includes/_configure.tpl
+++ b/chart/templates/developer-hub/includes/_configure.tpl
@@ -63,6 +63,13 @@
       done
       echo "OK"
 
+      echo -n "* Waiting for RHDH deployment: "
+      until [[ "$(kubectl -n ai-rhdh get deployment "backstage-${BACKSTAGE_CR_NAME}" -o=jsonpath='{.status.conditions[?(@.type=="Progressing")].reason}' --ignore-not-found)" == "NewReplicaSetAvailable" ]]; do
+        echo -n "_"
+        sleep 2
+      done
+      echo "OK"
+
       echo -n "* Patching RHDH Default App Config: "
       APPCONFIG_DATA=$(mktemp)
       RHDH_URL="https://$(kubectl get route -n "$NAMESPACE" "backstage-${BACKSTAGE_CR_NAME}" --ignore-not-found -o jsonpath={.spec.host})"
@@ -74,6 +81,13 @@
       echo -n "."
       kubectl patch configmap $APPCONFIG_CONFIGMAP -n $NAMESPACE --type='merge' -p="{\"data\":{\"app-config.base.yaml\":\"$(echo "$(cat ${APPCONFIG_DATA})" | sed 's/"/\\"/g' | sed 's/$/\\n/g' | tr -d '\n')\"}}"
 
+      echo "OK"
+
+      echo -n "* Waiting for RHDH deployment patch: "
+      until [[ "$(kubectl -n ai-rhdh get deployment "backstage-${BACKSTAGE_CR_NAME}" -o=jsonpath='{.status.conditions[?(@.type=="Progressing")].reason}' --ignore-not-found)" == "NewReplicaSetAvailable" ]]; do
+        echo -n "_"
+        sleep 2
+      done
       echo "OK"
 
       # Clean up temporary files


michael-valdron:
> Documentation (READMEs, Product Docs, Blogs, Education Modules, etc.)

N/A

chart/templates/developer-hub/includes/_configure.tpl
```diff
@@ -63,6 +63,13 @@
       done
       echo "OK"
 
+      echo -n "* Waiting for RHDH deployment: "
+      until [[ "$(kubectl -n ai-rhdh get deployment "backstage-${BACKSTAGE_CR_NAME}" -o=jsonpath='{.status.conditions[?(@.type=="Progressing")].reason}' --ignore-not-found)" == "NewReplicaSetAvailable" ]]; do
```
gabemontero, line 67: might want to look at `kubectl wait`  ... easier syntax wrt waiting on conditions to change

Jdubrick: +1, just looked at `kubectl wait` and it sounds perfect for this use case

michael-valdron: @gabemontero @Jdubrick I think the original intention for using `kubectl get` versus `kubectl wait` is for the output while waiting, does not look like there is a clean way to archive this with `kubectl wait`, I could replace all wait loops with `kubectl wait` just means you won't get anything printed until wait is completed for a given reason (condition met, error, timeout). Plus though is that we have more of a clean way to specify a timeout. wdyt?

michael-valdron: Could do this with `kubectl wait` to reproduce the same result:

```sh
echo -n "* Waiting for RHDH deployment: "
kubectl wait ... &
PROCESS_ID=$!

until ! ps -p $PROCESS_ID > /dev/null; do
    echo -n "_"
    sleep 2
done

wait $PROCESS_ID
if [ $? -ne 0 ]; then
    echo "FAIL"
    exit 1
fi
echo "OK"
```

Jdubrick: I think the `kubectl wait` is really clean and clear on what is happening, I do think if we go that route we should add a `timeout` flag as a failsafe

michael-valdron: > I think the `kubectl wait` is really clean and clear on what is happening

@Jdubrick In source yes, but in the logs there is no indication if the command is still waiting or terminated other than the result message and the output from the rest of the steps.

> I do think if we go that route we should add a `timeout` flag as a failsafe

I agree, the other wait steps in other parts needs this as well.

Jdubrick: In the spirit of having logs that are clear I'm totally fine with doing it the way we initial planned, I am thinking back to the task I'm currently working on for the auto disabling of the tls check for argocd and having clear logs is a necessity IMO. wdyt @gabemontero

michael-valdron: We could use [my suggestion](https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/41#discussion_r1956503905) to meet best of both worlds.

michael-valdron: @gabemontero @Jdubrick I've done some testing and seems that the current image used `registry.redhat.io/openshift4/ose-tools-rhel8:latest` that supports up to OCP 4.15 does not support the wait condition using `kubectl wait -n $NAMESPACE --for='jsonpath={.status.conditions[?(@.type=="Progressing")].reason}="NewReplicaSetAvailable"' <deployment/deployment-name> --timeout=<seconds>s`:
```
* Fetching yq: OK
* Waiting for 'backstages' CRD: OK
* Creating RHDH Extra Variables Secret: OK
* Creating RHDH Base App Config: OK
* Creating RHDH Dynamic Plugins Config: OK
* Creating RHDH instance: OK
* Waiting for RHDH route: OK
* Waiting for RHDH deployment: _error: jsonpath wait format must be --for=jsonpath='{.status.readyReplicas}'=3 or --for=jsonpath='{.status.readyReplicas}'
```

However changing to `registry.redhat.io/openshift4/ose-tools-rhel9` that supports OCP 4.16+ this command does work, with this being the case I wonder if I should go back to the wait loop I was using and include these changes under the migration to RHDH v1.4 changes?

Jdubrick: I think that is a good idea, revert to the loop we were previously using for now and then for the v1.4 changes we can update the image so it properly supports the OCP versions rhdh v1.4 is compatible with. We'll need to update that image anyways I am guessing since rhdh v1.4 goes higher than 4.15? @michael-valdron

michael-valdron: Correct, I just opened an issue to address this task: https://issues.redhat.com/browse/RHDHPAI-598

Also, I included the conversion from wait loops with `kubectl get` to `kubectl wait` as part of the scope of this issue.

michael-valdron: @gabemontero @Jdubrick Let me know if there is any further feedback on this, going to resolve this thread to address in https://issues.redhat.com/browse/RHDHPAI-598.

Jdubrick (approved):
lgtm with the rolled back changes to the old way using `until`.

chart/templates/developer-hub/includes/_configure.tpl
gabemontero: yep agree with the above plan @michael-valdron @Jdubrick sounds good ... simplify with wait when can work with 4.16+ (I was unaware it was that new)

PR #42, opened 2025-02-18
RHDHPAI-593: add option to allow API access for CI

jrichter1:
### What does this PR do?:
Adds the option to disable API auth to be used **for testing in CI only** using the `CI_DISABLE_AUTH` env variable.

This approach is taken from [RHTAP](https://github.com/redhat-appstudio/rhtap-cli/blob/9784e3b54f1517ace5c1f89122e2dd68bbf0f54f/installer/charts/rhtap-dh/templates/app-config-content.yaml#L67), allows the tests to run templates through unauthenticated API requests. 

https://issues.redhat.com/browse/RHDHPAI-591

### PR acceptance criteria:

- [x] Tested and Verified

  

### DIFF
--- a/scripts/configure-dh.sh
+++ b/scripts/configure-dh.sh
@@ -364,6 +364,12 @@ if [[ $LIGHTSPEED_INTEGRATION == "true" ]]; then
     fi
     echo -n "."
 fi
+# Only use for running end-to-end tests in CI, keep away from production
+if [[ $CI_DISABLE_AUTH == "true" ]]; then
+    EXTRA_APPCONFIG=$(echo "$EXTRA_APPCONFIG" | yq ".backend.auth.dangerouslyDisableDefaultAuthPolicy = true" -M -)
+    echo -n "."
+fi
+
 EXTRA_APPCONFIG=$EXTRA_APPCONFIG yq ".data[\"app-config.extra.yaml\"] = strenv(EXTRA_APPCONFIG)" $BASE_DIR/resources/developer-hub-app-config.yaml | \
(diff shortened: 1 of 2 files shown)


Jdubrick (approved):
lgtm

scripts/configure-dh.sh
```diff
@@ -364,6 +364,12 @@ if [[ $LIGHTSPEED_INTEGRATION == "true" ]]; then
     fi
     echo -n "."
 fi
+# Only use for running end-to-end tests in CI, keep away from production
+if [[ $CI_DISABLE_AUTH == "true" ]]; then
```
michael-valdron: Thinking for this variable a declaration to follow the others that defaults to `false` and a generalized name, such as `RHDH_DISABLE_AUTH_POLICY`, as there could be other reasons to disable the authentication policy:

```suggestion
if [[ $RHDH_DISABLE_AUTH_POLICY == "true" ]]; then
```

michael-valdron (changes requested):
Just the minor changes suggested and would be good to include this addition in the documentation somewhere. Perhaps a small section under [EXTRA_CONFIG.md](https://github.com/redhat-ai-dev/ai-rhdh-installer/blob/0542e44be17653cdebfe6a18ed253acc6edec62b/docs/EXTRA-CONFIG.md) explaining what this variable does as it seems to fit this documents scope of topics.

scripts/configure-dh.sh
jrichter1: Renamed the variable as suggested, added `false` as default, and a snippet in the doc for explanation.

michael-valdron (approved):

PR #44, opened 2025-02-24
[RHDHPAI-523] Add Flag To Prevent Helm Uninstall Crash Loop

Jdubrick:
### What does this PR do?:

This PR adds the `--ignore-not-found` flag to `kubectl delete` which allows the command to handle non-existent resources as successful deletions. This should eliminate the crash loop that is occurring during `Helm uninstall` where the resources are partially installed as the script won't error out if a resource is missing.

https://issues.redhat.com/browse/RHDHPAI-523

### PR acceptance criteria:

- [ ] Tested and Verified

  

Was having issues reproducing using the steps outlined in the linked JIRA above so to work around this I had to allow the install to complete and then manually go in and remove one of the resources (app configmag, env secret, etc) so that something was "missing/partial" before running the uninstall portion.

If anyone was able to properly reproduce with the instructions in JIRA and can let me know how this fix handles it I would very much appreciate that.

### DIFF
--- a/chart/templates/developer-hub/includes/_unconfigure.tpl
+++ b/chart/templates/developer-hub/includes/_unconfigure.tpl
@@ -17,25 +17,25 @@
       NAMESPACE="{{.Release.Namespace}}"
 
       echo -n "* Deleting RHDH instance: "
-      cat <<EOF | kubectl delete -n "$NAMESPACE" -f - >/dev/null
+      cat <<EOF | kubectl delete -n "$NAMESPACE" --ignore-not-found -f - >/dev/null
       {{ include "rhdh.include.backstage" . | indent 6 }}
       EOF
       echo "OK"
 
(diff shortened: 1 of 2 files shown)


michael-valdron (commented):
I think it might be better to ignore not found errors rather than removing `set -o errexit` all together, such as using `--ignore-not-found` on `kubectl delete` statements. @Jdubrick wdyt?

thepetk (commented):
+1 on what @michael-valdron mentions. I feel too is better to add the `--ignore-not-found` flag.

Jdubrick:
I didn't even see that flag when I was looking :( I will implement that! Thanks guys @michael-valdron @thepetk

Jdubrick:
Updated the PR to reflect the new changes and added the proper flag. fyi @thepetk @michael-valdron

michael-valdron (approved):

PR #49, opened 2025-03-10
[RHDHPAI-598] Bump OSE Tools Image

Jdubrick:
### What does this PR do?:

This PR contains the following changes:
- Bumps ose-tools to `rhel9` version `v.18`, this should provide support for up to OCP v4.18 without issue, and was tested on ocp v4.16 to maintain consistency with the docs.
- Updates `until kubectl get ...` logic with `kubectl wait` for appropriate resources.
  - `until kubectl get route ...` logic was kept in place as currently `kubectl wait` does not properly interact with Routes as they sit ontop of Ingresses and can be inconsistent.
- Refactored some pieces to ensure consistency throughout the configure templates.

https://issues.redhat.com/browse/RHDHPAI-598

### PR acceptance criteria:

- [x] Tested and Verified

  

Install like normal and observe expected functionality of the installer + template deployment

### DIFF
--- a/chart/templates/developer-hub/includes/_configure.tpl
+++ b/chart/templates/developer-hub/includes/_configure.tpl
@@ -1,7 +1,7 @@
 {{ define "rhdh.developer-hub.configure" }}
 {{ if (index .Values "developer-hub") }}
 - name: configure-developer-hub
-  image: "registry.redhat.io/openshift4/ose-tools-rhel8:latest"
+  image: "registry.redhat.io/openshift4/ose-tools-rhel9:v4.18.0-202502260503.p0.geb9bc9b.assembly.stream.el9"
   workingDir: /tmp
   command:
     - /bin/sh
--- a/chart/templates/developer-hub/includes/_configure.tpl
+++ b/chart/templates/developer-hub/includes/_configure.tpl
@@ -18,8 +18,7 @@
       CRD="backstages"
       echo -n "* Waiting for '$CRD' CRD: "
       while [ $(kubectl api-resources | grep -c "^$CRD ") = "0" ] ; do
-        echo -n "_"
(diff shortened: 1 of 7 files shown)


thepetk (approved):
lgtm.

My tests were successful and I was able to install RHDH and deploy a chatbot ST.

One thing I'd like to capture is

image

The truth is I got a timeout error, so most probably the error in topology comes from there. Both pods' logs are ok too.

Jdubrick:
@thepetk were you able to get RAG deployed? I was running into an issue with the chromadb setup so I have been looking into that but maybe it's just my side?

The timeout on the configure shouldn't be related, most likely an Operator setup hanging and you could try adding `--timeout=20m` for example to give it more breathing room

Jdubrick:
> lgtm.
> 
> My tests were successful and I was able to install RHDH and deploy a chatbot ST.
> 
> One thing I'd like to capture is
> 
> image
> 
> The truth is I got a timeout error, so most probably the error in topology comes from there. Both pods' logs are ok too.

@thepetk An Operator probably was slow to spin up which eats a lot of the base timeout, you could try adding `--timeout=15m` to rectify this

thepetk:
> > lgtm.
> > My tests were successful and I was able to install RHDH and deploy a chatbot ST.
> > One thing I'd like to capture is
> > image
> > The truth is I got a timeout error, so most probably the error in topology comes from there. Both pods' logs are ok too.
> 
> @thepetk An Operator probably was slow to spin up which eats a lot of the base timeout, you could try adding `--timeout=15m` to rectify this

Yeah +1, btw I'd like to capture it only in terms of the topology (not really related with the PR). It feels weird why the job seems to have failed one time but all the runs were completed with success.

PR #54, opened 2025-08-18
[RHDHPAI 622] adding dynamic remote cluster setup

JslYoon:
### What does this PR do?:
Added dynamic setup when bringing your own cluster

[RHDHPAI 622](https://issues.redhat.com/browse/RHDHPAI-622)

### PR acceptance criteria:
Added dynamic configuration of remote cluster(s), using configure.sh step, added documentations.

- [o] Tested and Verified

  

- [o] Documentation (READMEs, Product Docs, Blogs, Education Modules, etc.)

   

after helm install, run configure.sh

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,3 +74,13 @@ bash ./configure.sh
 **Note**: If you changed the installation namespace used by the installer you will first need to run `export NAMESPACE=<namespace used>` as the default value is `ai-rhdh`.
 
 For information related to extra configuration options, see [`docs/EXTRA-CONFIG.md`](./docs/EXTRA-CONFIG.md).
+
+## Remote Cluster Configuration
+
+- Refer to [cluster configuration docmentation](./docs/REMOTE-CLUSTER-CONFIG.md) for information about how to configure remote cluster.
+
+The installer supports connecting to multiple Kubernetes clusters for multi-cluster visibility of resources, pipelines, and applications. During configuration, you can add remote clusters that RHDH will connect to for:
+
+- Viewing Kubernetes resources across clusters
+- Tracking ArgoCD applications deployed to remote clusters
+- Using software templates that deploy to various target clusters
--- a/configure.sh
+++ b/configure.sh
@@ -37,6 +37,10 @@ export QUAY__DOCKERCONFIGJSON=${QUAY__DOCKERCONFIGJSON:-''}
 export QUAY__API_TOKEN=${QUAY__API_TOKEN:-''}
 export LIGHTSPEED_MODEL_URL=${LIGHTSPEED_MODEL_URL:-''}
 export LIGHTSPEED_API_TOKEN=${LIGHTSPEED_API_TOKEN:-''}
+export REMOTE_K8S_SA=${REMOTE_K8S_SA:-''}
+export REMOTE_K8S_SA_TOKEN=${REMOTE_K8S_SA_TOKEN:-''}
+export REMOTE_K8S_URL=${REMOTE_K8S_URL:-''}
+export REMOTE_K8S_AUTH_PROVIDER=${REMOTE_K8S_AUTH_PROVIDER:-'serviceAccount'}
 
 # Skipped optional variables
 export BYPASS_OPTIONAL_INPUT=''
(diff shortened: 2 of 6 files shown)


configure.sh
```diff
@@ -16,6 +16,7 @@ export RHDH_GITHUB_INTEGRATION=${RHDH_GITHUB_INTEGRATION:-true}
 export RHDH_GITLAB_INTEGRATION=${RHDH_GITLAB_INTEGRATION:-false}
 export LIGHTSPEED_INTEGRATION=${LIGHTSPEED_INTEGRATION:-false}
 export RHDH_SIGNIN_PROVIDER=${RHDH_SIGNIN_PROVIDER:-''}
+export REMOTE_CLUSTER_ENABLED=${REMOTE_CLUSTER_ENABLED:-false}
```
Jdubrick: Am I missing where we actually use this var other than setting it in this script based on what the cluster count is? It looks like we use the cluster count as the source of enablement making this var unused?

JslYoon: Made it for bookkeeping purpose, but we don't technically need it. Should I get rid of it? @Jdubrick

Jdubrick: I would yeah

configure.sh
```diff
@@ -214,6 +218,67 @@ if [ -z "${QUAY__DOCKERCONFIGJSON}" ]; then
...
+        if [ -z "${cluster_token}" ]; then
+            echo "Service account token is required, try again."
+        fi
+    done
+    
+    # Optional auth provider
+    read -p "Enter auth provider (default: serviceAccount): " cluster_auth_provider
+    if [ -z "${cluster_auth_provider}" ]; then
+        cluster_auth_provider="serviceAccount"
+    fi
```
Jdubrick: We should make this one a environment variable they can set if they desire (removing it from the CLI portion of the script) since everything else is referencing service accounts I don't know if we want them to be able to put something here. Have we confirmed if they do put something else here if it will work?

docs/REMOTE-CLUSTER-CONFIG.md
```diff
@@ -0,0 +1,148 @@
...
+
+# Grant view permissions across the cluster
+kubectl create clusterrolebinding rhdh-remote-access \
+  --clusterrole=view \
+  --serviceaccount=default:rhdh-remote-access
+
+# For OpenShift: Also grant access to routes
+kubectl create clusterrolebinding rhdh-remote-access-routes \
+  --clusterrole=admin \
+  --serviceaccount=default:rhdh-remote-access
```
maysunfaisal: could you explain what exactly this is needed for?

docs/REMOTE-CLUSTER-CONFIG.md
```diff
@@ -0,0 +1,148 @@
...
+EOF
+
+# Wait a moment for token to be generated, then retrieve it
+kubectl get secret rhdh-remote-access-token -o jsonpath='{.data.token}' | base64 -d
+```
+
+For **Kubernetes 1.23 and earlier**:
+```bash
+# Tokens are automatically created, find the secret name
+kubectl get serviceaccount rhdh-remote-access -o jsonpath='{.secrets[0].name}'
```
maysunfaisal: I think you can remove this line, since you have the exact same line next step

README.md
```diff
@@ -74,3 +74,52 @@ bash ./configure.sh
 **Note**: If you changed the installation namespace used by the installer you will first need to run `export NAMESPACE=<namespace used>` as the default value is `ai-rhdh`.
 
 For information related to extra configuration options, see [`docs/EXTRA-CONFIG.md`](./docs/EXTRA-CONFIG.md).
+
+## Remote Cluster Configuration
```
maysunfaisal, line 78: I vote to keep this info out of the main README since you already have a nice detailed doc in docs/REMOTE-CLUSTER-CONFIG.md,

maybe under this header you can just say - refer to docs/REMOTE-CLUSTER-CONFIG.md for information about how to configure remote cluster

configure.sh
```diff
@@ -214,6 +218,67 @@ if [ -z "${QUAY__DOCKERCONFIGJSON}" ]; then
     BYPASS_OPTIONAL_INPUT+=",QUAY__DOCKERCONFIGJSON"
 fi
 
+# Configure remote clusters
```
maysunfaisal, line 221: I would say also gate all this logic behind a flag like `ENABLE_REMOTE_CLUSTER_SETUP`? It can be false by default

or maybe gate it with a question like would you like to enable remote cluster setup? Y/N

Maybe I don't want to do any remote cluster setup, so I dont want to answer these questions.. 🤔

configure.sh
JslYoon: Yeah that makes sense, don't really see a reason someone would use something other than the service account, I made it into an env variable.

configure.sh
JslYoon: got rid of it

docs/REMOTE-CLUSTER-CONFIG.md
JslYoon: Removed

docs/REMOTE-CLUSTER-CONFIG.md
JslYoon: I thought route access was needed to create an instance on the cluster no? I can delete it.

README.md
JslYoon: Changed!

configure.sh
JslYoon: I gated it with y/n!

README.md
```diff
@@ -74,3 +74,54 @@ bash ./configure.sh
...
+
+- Refer to [cluster configuration docmentation](./docs/REMOTE-CLUSTER-CONFIG.md) for information about how to configure remote cluster
```
Jdubrick:
```suggestion
- Refer to [cluster configuration docmentation](./docs/REMOTE-CLUSTER-CONFIG.md) for information about how to configure remote cluster.
```

README.md
```diff
@@ -74,3 +74,54 @@ bash ./configure.sh
...
+   
+   === Remote Cluster 1 ===
+   Enter remote cluster service account name (CTRL-D to finish): rhdh-remote-access
+   Enter remote cluster URL: https://api.prod.example.com:6443
+   Enter remote cluster service account token: your-service-account-token
+   Remote cluster 1 configured: rhdh-remote-access (https://api.prod.example.com:6443)
+   
+   [Press CTRL-D to finish adding clusters]
+   ```
+
```
Jdubrick: Do we need this section if we are deferring to the `REMOTE-CLUSTER-CONFIG` doc?

docs/REMOTE-CLUSTER-CONFIG.md
maysunfaisal: I am not sure either, can you verify?

README.md
maysunfaisal: same thoughts https://github.com/redhat-ai-dev/ai-rhdh-installer/pull/54#discussion_r2289473659

JslYoon: We don't need them, they're redundant. Just deleted the content

README.md
JslYoon: resolved!

docs/REMOTE-CLUSTER-CONFIG.md
JslYoon: Just checked, it's url view only, so I think it's safe to delete

chart/values.yaml
```diff
@@ -18,7 +18,7 @@ openshift-gitops:
           cpu: "1"
           memory: 3Gi
   # TLS check can cause unexpected hanging during login: https://argo-cd.readthedocs.io/en/stable/user-guide/commands/argocd_login/#options
-  skip-test-tls: false
+  skip-test-tls: true
```
maysunfaisal: probably revert

dynamic-plugins/tekton-plugins.yaml
```diff
@@ -30,6 +30,27 @@ plugins:
...
+              - authProvider: ${REMOTE_K8S_AUTH_PROVIDER_1}
+                name: ${REMOTE_K8S_SA_1}
+                serviceAccountToken: ${REMOTE_K8S_SA_TOKEN_1}
+                skipTLSVerify: true
+                url: ${REMOTE_K8S_URL_1}
+              - authProvider: ${REMOTE_K8S_AUTH_PROVIDER_2}
+                name: ${REMOTE_K8S_SA_2}
+                serviceAccountToken: ${REMOTE_K8S_SA_TOKEN_2}
+                skipTLSVerify: true
+                url: ${REMOTE_K8S_URL_2}
```
maysunfaisal: Should these be committed?

chart/values.yaml
JslYoon: Yep my bad, for got to change them back

dynamic-plugins/tekton-plugins.yaml
JslYoon: They shouldn't... deleted them, and also made it so that the cluster information doesn't overlap on top of each other per configuration

maysunfaisal (approved):
lgtm

Jdubrick (approved):
lgtm

PR #55, opened 2025-08-28
fixing bug when secrets were failing when remote k8s cluster was set twicd

JslYoon:
fixing bug when secrets were failing when remote k8s cluster was set twicd


catalogs.yaml
```diff
@@ -1,3 +1,4 @@
 # List of Software Template catalogs to setup with RHDH
 catalogs:
-  - https://github.com/redhat-ai-dev/ai-lab-template/blob/main/all.yaml
+  # - https://github.com/redhat-ai-dev/ai-lab-template/blob/main/all.yaml
+  - https://github.com/JslYoon/ai-lab-template/tree/JslYoon/RHDHPAI-581/all.yaml
```
Jdubrick:
```suggestion
  - https://github.com/redhat-ai-dev/ai-lab-template/blob/main/all.yaml
```

chart/values.yaml
```diff
@@ -18,7 +18,7 @@ openshift-gitops:
           cpu: "1"
           memory: 3Gi
   # TLS check can cause unexpected hanging during login: https://argo-cd.readthedocs.io/en/stable/user-guide/commands/argocd_login/#options
-  skip-test-tls: false
+  skip-test-tls: true
```
Jdubrick:
```suggestion
  skip-test-tls: false
```

scripts/configure-dh.sh
```diff
@@ -427,6 +405,8 @@ configure_dh() {
         check_status || return 1
     fi
 
+    # Defer patching dynamic plugins until after optional remote cluster processing
+
```
Jdubrick: We can probably drop this comment

Jdubrick (commented):
Quick couple cleanup comments

scripts/configure-dh.sh
JslYoon: dropped!

catalogs.yaml
JslYoon: Oops always forgetting this :( just fixed!

chart/values.yaml
JslYoon: changed!

Jdubrick (approved):
lgtm, let's wait to merge until we're back Tuesday @JslYoon

